PR #137, opened 2026-06-22
fix(skills): confine catalog paths with filepath.IsLocal (CodeQL path-injection, IRO-85)

omerzamir:
## Summary
Fixes the 4 HIGH `go/path-injection` CodeQL alerts in the skills host (parent IRO-82, IRO-85):
- `internal/host/skills/catalog.go:93,96` — alerts #22, #23
- `internal/host/skills/source.go:254,258` — alerts #24, #25

The user-supplied `name`/`version` were already charset-validated (`validName`/`validVersion`) and confined by the bespoke `withinRoot` helper, but **CodeQL does not recognise those as sanitizers**, so the dataflow stayed tainted into `os.ReadFile`/`os.Stat`/`os.RemoveAll`.

## Fix
Route the components through **`filepath.IsLocal`** (Go 1.20+) — the stdlib's lexical path-confinement predicate, which CodeQL **does** recognise as a path-injection barrier — applied to the relative path *before* it is joined onto the root. `IsLocal` rejects absolute paths, the empty path, and any `../` escape, so the path reaching the filesystem is provably confined. The bespoke `withinRoot` helper is removed in favour of the stdlib check (charset validators kept for clear errors + defence-in-depth).

This is the *refactor-so-the-taint-is-provably-broken* option from the acceptance criteria, not an alert dismissal.

## Tests
- `TestDirSourceOpenRejectsTraversal` — `../`, absolute-path, and dot-segment payloads all rejected; an out-of-root secret bundle is never read.
- `TestCatalogRemoveRejectsTraversal` — same payloads on the destructive path; an out-of-root sentinel survives.
- `TestDirSourceOpenSymlinkBoundary` — documents the agent-facing symlink boundary.

## Verification
- `CGO_ENABLED=1 go test ./internal/host/skills/` → 40 pass (+3 new)
- `CGO_ENABLED=1 go test ./internal/host/api/ ./internal/host/isolation/` → 128 pass
- `go vet ./internal/host/skills/` → clean

## Security lenses
- **Trust boundary integrity** — narrows, never widens, what the skills host will read/delete from agent-supplied input. No queue/sandbox capability change.
- **Mandatory gateway** — install remains a separate gateway-gated step; this only hardens catalog read/remove.

Re-run CodeQL on merge to confirm 0 open `go/path-injection` alerts. Hand to @QA for verification.

### DIFF
--- a/internal/host/skills/catalog.go
+++ b/internal/host/skills/catalog.go
@@ -80,16 +80,21 @@ func (d DirSource) Remove(name, version string) error {
 	if !validName(name) {
 		return fmt.Errorf("skills: invalid skill name %q", name)
 	}
-	target := filepath.Join(d.Root, name)
+	rel := name
 	if version != "" {
 		if !validVersion(version) {
 			return fmt.Errorf("skills: invalid skill version %q", version)
 		}
-		target = filepath.Join(target, version)
+		rel = filepath.Join(name, version)
 	}
-	if !withinRoot(d.Root, target) {
+	// Confine the user-supplied relative path with filepath.IsLocal before joining
+	// onto the root, so the path handed to os.Stat/os.RemoveAll provably cannot
+	// escape the catalog. See DirSource.Open for the rationale; IsLocal rejects
+	// absolute paths and any "../" escape and is the barrier the analyser tracks.
+	if !filepath.IsLocal(rel) {
 		return fmt.Errorf("skills: resolved path escapes the catalog root")
 	}
+	target := filepath.Join(d.Root, rel)
 	if _, err := os.Stat(target); err != nil {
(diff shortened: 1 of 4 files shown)


omerzamir:
Superseded by #135 (IRO-90), which merged to `main` at 21:34 UTC and fixes the identical 4 `go/path-injection` alerts (#22–#25) via the `resolveBundlePath` barrier (cleaned-path containment, data-flow-visible to CodeQL) plus traversal unit tests. This PR's `filepath.IsLocal` approach is an equivalent second fix and now conflicts because IRO-90 rewrote the same lines — merging it would re-churn identical code with no security delta. Closing as duplicate; the underlying vulnerability is fixed and tested on `main`. QA verification continues on [IRO-85].

PR #182, opened 2026-06-24
ci(release): cross-machine bit-reproducibility via digest-pinned Linux build container (IRO-127)

omerzamir:
## What & why

`reproducibility.yml` only proved **same-runner** reproducibility — build twice on one
runner, assert the hashes match. But the absolute hashes drift across runners (IRO-127):
the cgo C toolchain (gcc/glibc) that compiles the vendored SQLCipher amalgamation varies
by `ubuntu-latest` image patch, so an **independent third party rebuilding from source does
not reproduce the published `SHA256SUMS`** — the stronger guarantee reproducible-builds
reviewers and OpenSSF Scorecard actually want.

## Changes

- **`release.yml`** — the `linux/amd64` + `linux/arm64` build legs now run inside a
  **digest-pinned** `golang:1.23.12-bookworm` container, so the Go compiler/linker **and**
  gcc/glibc are byte-identical for everyone. macOS/Windows still build natively on the host.
- **`release.yml`** — Linux archives are packaged **deterministically**
  (`SOURCE_DATE_EPOCH` from the commit, `tar --sort=name --owner=0 --group=0
  --numeric-owner`, `gzip -n`) so the **linux rows of `SHA256SUMS`** — not just the binaries
  inside — reproduce. (A plain `tar -czf` bakes in mtimes + a gzip timestamp.)
- **`reproducibility.yml`** — builds twice in the pinned container on **two different host
  images** (`ubuntu-22.04` + `ubuntu-24.04`) and fails unless the binaries **and** the
  deterministic archive are byte-identical **across hosts**. Real cross-machine assertion.
- **`docs/release-runbook.md`** — documented independent-rebuild procedure (exact image
  digest + copy-pasteable build command) that reproduces the published `linux/amd64` set.

## Verification

Run locally under Docker (`--platform linux/amd64` in the pinned image):

- Two cold builds → byte-identical `controlplane`, `ironctl`, `sandbox`.
- Deterministic archive of each → byte-for-byte identical (`99be368c…`).
- `actionlint` clean on both workflows.

This PR's own `Reproducibility` run is the live cross-machine (two-host) proof in real CI.

## Supply-chain lenses

- **Reproducibility** — primary: published Linux artifacts re-derivable bit-for-bit from a
  known commit + documented container.
- **Pinned, auditable actions** — builder image pinned by `@sha256` digest, in sync across
  `release.yml` / `reproducibility.yml` / the runbook.
- **Build matrix fidelity** — matrix unchanged (macOS amd64+arm64, Linux amd64+arm64,
  Windows amd64; CGO, Go 1.23+); only the Linux build *environment* changed.
- **Least-privilege CI** — no token-scope changes.

No changes to signing, attestation, required-checks, or installer verification paths.

Closes IRO-127.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,14 @@ name: Release
 # amalgamation), so each target is built on a runner with a matching C toolchain
 # rather than cross-compiled blind. The only cross-build is linux/arm64, which is
 # pure C and links cleanly with the aarch64 GNU cross-compiler.
+#
+# The Linux legs build INSIDE a digest-pinned container (golang:1.23.12-bookworm,
+# see BUILDER_IMAGE below) so the cgo C toolchain (gcc + glibc) that compiles the
+# SQLCipher amalgamation is byte-identical for everyone — not whatever patch the
+# ubuntu-latest image happens to ship that day. That is what makes the published
+# linux/amd64 + linux/arm64 binaries reproducible by an INDEPENDENT third party,
+# not just twice on the same runner (IRO-127). macOS/Windows build natively on the
+# host. The exact rebuild procedure is documented in docs/release-runbook.md.
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


omerzamir:
Superseded by #214 and closing.

This branch was cut before IRO-192 (#211), IRO-193 (#208), and IRO-195 (#209) landed on `main`. Merging it as-is would silently **revert** all three:
- the `cosign-installer` 2.6.3 pin → re-breaks `sign-blob` (`.sig`/`.pem`),
- the `formula:` job + `paths-ignore` → drops Homebrew auto-tracking,
- the **bullseye** glibc-2.31 floor + objdump guard → binaries stop running on Ubuntu 22.04 / Debian 12 / RHEL 9 (it pins **bookworm**).

The still-wanted IRO-127 deliverable — the two-host cross-machine reproducibility check + the from-source runbook section — has been rebased onto current `main` and retargeted to the `golang:1.23.12-bullseye` container the release actually ships, in #214. CEO approval (confirmation 612d9e2e) carries forward; the deliverable is unchanged, only its base is corrected.

PR #209, opened 2026-06-25
ci(release): pin cosign to 2.6.3 to fix sign-blob crash (IRO-195)

omerzamir:
## Launch blocker fix — IRO-195

The **Release** workflow has failed on every recent tag (v0.1.88–v0.1.91) at the
cosign **"Sign the checksum file"** step:

```
Error: signing dist/SHA256SUMS: create bundle file: open : no such file or directory
```

### Root cause
`sigstore/cosign-installer` was used **without a `cosign-release` pin**, so it pulled
**cosign 3.x**. cosign 3.0.0 flipped `sign-blob` to default to the new Sigstore bundle
format, made `--bundle` **required**, and deprecated/ignored
`--output-signature`/`--output-certificate`. With no `--bundle` path, cosign tried to
write a bundle to an empty path and crashed. Because the step runs under
`set -euo pipefail`, the job aborted **before** the build-provenance attestation step →

- no `SHA256SUMS.sig` / `SHA256SUMS.pem` on any release,
- `gh attestation verify <tarball>` → **HTTP 404** (attest step never ran),
- post-release install **smoke jobs skipped** (they `need` the release job).

Regression of **IRO-46** (SLSA provenance) and **IRO-32** (WS-A supply-chain trust).

### Fix
Pin `cosign-release: v2.6.3` (latest 2.x) on **both** cosign-installer steps (the
release/sign job and the smoke/verify job). In 2.6.x the new bundle format is **opt-in**
(`--new-bundle-format=true`), so `sign-blob --output-signature/--output-certificate`
still emits the detached `.sig`/`.pem` that the anonymous `cosign verify-blob` smoke
step, the README "Verify the download", `docs/security.md`, and `docs/release-runbook.md`
all consume — **no command or doc churn**. Pinning also makes the signing tool a
deterministic, reproducible dependency (consistent with IRO-92 pinned-deps posture).

### Verification
- `release.yml` parses as valid YAML; diff is exactly the two `cosign-release: v2.6.3` pins.
- cosign 2.6.x release notes confirm new-bundle-format is opt-in there; the default flip
  to required-bundle landed in v3.0.0. `cosign-linux-amd64` asset exists for v2.6.3.
- **End-to-end validation happens on merge:** pushing to `main` cuts a fresh tag and runs
  the full Release workflow. Success criteria: green Release run, `.sig`/`.pem` attached,
  `gh attestation verify` passes, and the cosign-verify + install smoke jobs pass. Then
  re-run IRO-190 pre-launch smoke (checksum already PASS; need attestation + cosign verify).

Scope confined to `release.yml`; no other workflow uses cosign-installer.

Security lenses: supply-chain / release integrity (restores signing + provenance);
no trust boundary widened — strengthens determinism. Trust-model/release-gating → CEO review.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -292,7 +292,16 @@ jobs:
             -o "cyclonedx-json=ironclaw_${VERSION}.cdx.json"
           gh release upload "${TAG}" "ironclaw_${VERSION}.spdx.json" "ironclaw_${VERSION}.cdx.json" --clobber
 
+      # Pin cosign to the 2.x line. cosign 3.x flips `sign-blob` to the new Sigstore
+      # bundle format by default, makes `--bundle` required, and deprecates/ignores
+      # `--output-signature`/`--output-certificate` — which broke this step (it tried
+      # to write a bundle to an empty path → "create bundle file: open :"). 2.6.3 keeps
+      # the detached `.sig`/`.pem` output that the verify smoke step, README "Verify the
+      # download", docs/security.md, and the release runbook all rely on. Pinning also
+      # makes the signing tool a deterministic, reproducible dependency. See IRO-195.
       - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+        with:
+          cosign-release: v2.6.3
 
       - name: Sign the checksum file (cosign keyless) and attach it
         env:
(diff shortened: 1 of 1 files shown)


omerzamir:
## Verification for the reviewer/merger — pin is correct, safe to land on green

Confirming the fix is sound before merge:

- **`cosign-release: v2.6.3` is a real tag and the latest 2.x release** (`sigstore/cosign` 2.6.3, 2026-04-06) — checked against the upstream release list. The 2.x line keeps `sign-blob` emitting **detached** `SHA256SUMS.sig` / `.pem` via `--output-signature`/`--output-certificate`. The bundle-format default that broke the job (`create bundle file: open :`) is **3.x-only**, so pinning 2.6.3 restores the exact flow the README "Verify the download", `docs/security.md`, and the release runbook document.
- **Both** installer steps (sign **and** verify-blob) are pinned to the same 2.6.3, so signing and verification agree.
- This is a **regression fix** (re-pinning a tool to its known-good major), not a trust-model change — no supply-chain boundary is widened; it restores IRO-46 provenance + IRO-32 verify behavior.

**Why one pin fixes all three reported symptoms:** the sign step no longer aborts → the `release` job completes → `attest-build-provenance` runs (fixes the `gh attestation verify` 404) → the `smoke`/`smoke_windows` jobs (`needs: [version, release]`) stop being skipped. The merge's own `main` push cuts the next `v0.1.x` release through the fixed workflow, which is the live end-to-end proof.

Once required checks are green this is safe to squash-merge. I'm standing by as a merge backstop if needed.

PR #213, opened 2026-06-26
ci(release): build Linux archives as static musl — fix GLIBC_2.38 floor (IRO-192)

omerzamir:
## Problem (launch blocker, IRO-192)

The prebuilt Linux release binaries were **dynamically linked against glibc** and inherited the build host's GLIBC symbol versions. Built on `ubuntu-latest` (glibc 2.39), they demanded `GLIBC_2.38` and failed to even print `--version` on every mainstream server LTS:

| Distro | glibc | Before |
|---|---|---|
| Ubuntu 22.04 | 2.35 | ❌ |
| Debian 12 | 2.36 | ❌ |
| RHEL/Rocky/Alma 9 | 2.34 | ❌ |
| Amazon Linux 2023 | 2.34 | ❌ |

Linux + gVisor is the primary production target and the README headline install is `curl … | sh`, so a brand-new Linux user on a normal server got a binary that wouldn't start.

## Root cause

`CGO_ENABLED=1` (the SQLCipher binding) makes the Go binary dynamically link the build host's glibc. Crucially, `go-sqlcipher` uses the **vendored libtomcrypt** crypto backend (`-DSQLCIPHER_CRYPTO_LIBTOMCRYPT`) — confirmed in the dependency source — so the only external link is **libc itself** (no OpenSSL). That makes a fully static link viable.

## Fix

Build the Linux archives as **fully static musl** binaries:
- `CC=musl-gcc` (`apt: musl-tools`), `-linkmode external -extldflags '-static -Wl,--build-id=none'`.
- Each Linux arch builds **natively** — `amd64` on `ubuntu-latest`, `arm64` on `ubuntu-24.04-arm` — replacing the glibc `aarch64-linux-gnu` cross-compile.
- The reproducible-build `--build-id=none` is preserved (folded into the single `-extldflags` value).

A static musl binary has **no libc floor at all** — it runs on any Linux / any libc.

## Regression guard (so this can't silently come back)

New **`smoke_linux_compat`** job re-runs the issue's exact repro on every release: installs the just-published archive via the real checksum-verifying `scripts/install.sh` **inside old-glibc containers** and asserts `ironctl version` + `ironctl doctor` execute:
- `debian:bookworm-slim` (glibc 2.36, the issue repro), `ubuntu:22.04` (2.35), `almalinux:9` (2.34 — RHEL/AL2023 floor) on amd64;
- `debian:bookworm-slim` on arm64.

Fail-closed like the other smoke jobs (a red run blocks the GHCR image via `image.yml`).

## Reproducibility kept in sync

`reproducibility.yml` now rebuilds with the **same** musl-static toolchain, so the determinism check verifies the actually-shipped artifact rather than a glibc binary that is no longer released.

## Verification
- Both workflows YAML-parse clean; jobs: version, build, release, smoke, **smoke_linux_compat**, smoke_windows, formula.
- Reasoning confirmed against `go-sqlcipher@v4.4.2` source (libtomcrypt, no OpenSSL link).
- **Authoritative validation is in-CI**: `reproducibility.yml` runs on this PR (proves the musl-static build is bit-for-bit deterministic); `smoke_linux_compat` runs on the next release (proves the binary runs on the old-glibc floor distros). Local Docker validation was unavailable in the build sandbox this run.

Closes IRO-192.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,8 +6,20 @@ name: Release
 #
 # CGO is mandatory (the encrypted-SQLite binding compiles the vendored SQLCipher C
 # amalgamation), so each target is built on a runner with a matching C toolchain
-# rather than cross-compiled blind. The only cross-build is linux/arm64, which is
-# pure C and links cleanly with the aarch64 GNU cross-compiler.
+# rather than cross-compiled blind.
+#
+# The Linux archives are built as **fully static musl** binaries (CC=musl-gcc,
+# -extldflags=-static). A dynamically-linked glibc cgo binary bakes in the *build
+# host's* GLIBC symbol versions: building on ubuntu-latest (glibc 2.39) made the
+# binary demand GLIBC_2.38 and fail to even print --version on Ubuntu 22.04 /
+# Debian 12 / RHEL 9 / Amazon Linux 2023 — every mainstream server LTS (IRO-192).
+# go-sqlcipher uses the vendored libtomcrypt crypto backend
+# (-DSQLCIPHER_CRYPTO_LIBTOMCRYPT), so the only external link is libc itself; a
+# static musl link drops the glibc floor entirely and runs on any Linux / any libc.
+# Each Linux arch builds natively (amd64 on x86_64, arm64 on an arm64 runner) so
+# musl-gcc is the native toolchain — never a blind cross-compile. The
+# smoke_linux_compat job re-runs the IRO-192 repro on old-glibc distros every
+# release so this can never silently regress.
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


omerzamir:
Closing as redundant. IRO-192 was already fixed and merged on `main` via #211 (commit `e172c9d`), which builds the Linux binaries inside a pinned `golang:1.23.12-bullseye@sha256` container (glibc 2.31 floor — covers Ubuntu 22.04 / Debian 12 / RHEL 9 / Amazon Linux 2023, all ≥2.34>2.31) and adds an `objdump -T` glibc-floor assertion (≤2.31) in both `release.yml` and `ci.yml` so a regression reds every PR at build time.

This branch was inadvertently based on a commit *before* #211 merged, so it edited a stale `release.yml` and proposed a competing static-musl rewrite. The merged container approach already resolves the issue with a green, guarded build; replacing it with an unvalidated rewrite would be churn. No action needed.

PR #247, opened 2026-06-30
feat(examples): hello-ironclaw zero-cred end-to-end check + CI smoke test (IRO-240)

omerzamir:
## What

Adds **`examples/hello-ironclaw/`** — the single, self-contained, copy-runnable end-to-end example IRO-240 asks for: it proves IronClaw works with **zero credentials** and doubles as a **user-facing CI smoke test**.

```sh
examples/hello-ironclaw/run.sh        # build → up → assert reply → tear down
```

One command builds the sandbox image, brings up the offline demo control-plane, sends a chat through the **real secured path** (engage → per-session Docker sandbox → encrypted SQLCipher queue → delivery), **asserts** the offline `mock-agent` reply comes back, and tears the demo down. It exits non-zero (dumping the control-plane logs) if the round-trip ever breaks.

## Why

Docs explain features; a complete worked run a user can copy-paste delivers the "aha". We had scenario recipe fragments (IRO-115: `scheduled-report`, `webhook-responder`, `slack-triage`) but **no single end-to-end example that doubles as a smoke test and a marketing artifact** — and they print without asserting, so they can't fail a build. This fills exactly that gap.

## Contents

- `examples/hello-ironclaw/run.sh` — self-managed lifecycle, asserts the `mock-agent received: …` echo, `--keep` / `--attach` / `SKIP_BUILD` flags.
- `examples/hello-ironclaw/README.md` — exact commands, **expected output**, the real-provider swap note, and the demo's security relaxations (runc not gVisor; gateway/queues/credential-custody unchanged).
- `.github/workflows/example-smoke.yml` — **additive, non-gating** workflow that runs the exact user command on `ubuntu-latest` (Docker available). Complements the Go-level hermetic test (`internal/smoke/demo_smoke_test.go`, run by `go test`) by covering the **published HTTP/compose/runtime path** the unit test cannot reach. Pinned action SHA, `contents: read` only, path-filtered + `workflow_dispatch`.
- Cataloged in `examples/README.md` ("Start here") and linked from `docs/quickstart.md`.

## Supply-chain lenses touched

- **Least-privilege CI** — new workflow is `contents: read` only.
- **Pinned, auditable actions** — `actions/checkout` pinned to a SHA (matches `ci.yml`).
- **Fail loud, fail closed** — the script asserts and exits non-zero on a broken round-trip; the workflow is additive and **not** a required status check, so it can never block a release.

## Verification

- `bash -n`, `shellcheck` (clean), `--help` exercised.
- `actionlint` + YAML parse clean.
- Reply contract verified against source: `mockReplyPrefix = "mock-agent received: "` (`internal/sandbox/provider/mock.go`), routes in `internal/host/api/ui_chat.go`, demo token/group/seeding in `docker-compose.demo.yml`.
- **Live e2e**: Docker is unavailable on the authoring host, so the live Docker round-trip is verified by `example-smoke.yml` running in CI on this PR (it is path-filtered to trigger here).

Closes IRO-240.

### DIFF
--- a/.github/workflows/example-smoke.yml
+++ b/.github/workflows/example-smoke.yml
@@ -1,0 +1,56 @@
+name: Example smoke (zero-cred chat)
+
+# Hermetic, user-facing smoke test of the zero-credential demo path: it runs the
+# exact command a first-timer runs — examples/hello-ironclaw/run.sh — which builds
+# the sandbox image, brings up docker-compose.demo.yml, sends a chat through the
+# real HTTP route into a per-session Docker sandbox, and asserts the offline
+# mock-agent's reply comes back. The offline `mock` provider makes no network call,
+# so no credential is needed; ubuntu-latest ships Docker, jq, and curl.
+#
+# This complements the Go-level hermetic test (internal/smoke/demo_smoke_test.go,
+# run by CI's `go test`): that guards the queue<->loop<->mock contract; this guards
+# the published user path (compose file, demo entrypoint seeding, HTTP routes, the
+# Docker/runc sandbox runtime) that the unit test cannot reach.
+#
+# Additive and non-gating: it is NOT a required status check, so it can never block
+# a release — it surfaces a broken first-five-minutes before it reaches a user.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "examples/hello-ironclaw/**"
+      - "docker-compose.demo.yml"
+      - "container/**"
+      - "cmd/sandbox/**"
+      - "cmd/controlplane/**"
+      - "internal/host/api/ui_chat.go"
+      - ".github/workflows/example-smoke.yml"
+  pull_request:
+    paths:
+      - "examples/hello-ironclaw/**"
+      - "docker-compose.demo.yml"
+      - "container/**"
+      - "cmd/sandbox/**"
+      - "cmd/controlplane/**"
+      - "internal/host/api/ui_chat.go"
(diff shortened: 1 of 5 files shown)


omerzamir:
## Merge declined — `hello-ironclaw` smoke check is RED

CEO review before merge. The merge condition was "merge once the `hello-ironclaw` CI check is green." That check **failed** ([run 28470649147](https://github.com/IronSecCo/ironclaw/actions/runs/28470649147)), so I'm not merging — this is the exact engage → sandbox → reply path the example exists to demonstrate, and it can't ship red.

```
==> sending a chat message to 'mock-agent': "hello from hello-ironclaw 2293"
FAIL: no reply within 45s — the engage -> sandbox -> reply path is broken
==> tearing the demo down
```

Control-plane logs show the sandbox **did** launch (`host/session: launched sandbox for ses_bb166d1b0ba62cef`) but no reply arrived inside the 45s window. Everything else is green (build, CodeQL, reproducibility, WS-G, glibc-floor, CLA).

Routed back to @Relay (tracked as IRO-247) to diagnose — most likely a CI cold-start budget issue (fresh image build + container boot + mock-agent start can exceed 45s on `ubuntu-latest` vs ~2.6s locally), or confirming the reply rides the 2s delivery poll rather than the 60s sweep. Push the fix to this branch; once `hello-ironclaw` is green I'll merge via the reviewer-approve dispatch path.

PR #259, opened 2026-07-01
[DIAGNOSTIC — do not merge] IRO-253 hello-ironclaw delivery instrumentation run

omerzamir:
**Throwaway diagnostic PR — DO NOT MERGE. Close after reading the CI logs.**

Purpose: run `example-smoke.yml` (hello-ironclaw, Linux CI Docker) with the IRO-253 delivery logger present, to disambiguate the two hypotheses for the RED round-trip:

- `delivery: due outbound read ... due=0` (with a sandbox `wrote reply (N bytes)` above it) → **H2** (host reads 0 rows from the sandbox-written outbound.db across the bind mount).
- `delivery: handled outbound ... channel= platform=` empty → **H1b** (webchat adapter drops an empty-platform reply).

Composition = current `main` + the IRO-253 delivery logger (PR #258) + the hello-ironclaw example & smoke check cherry-picked from `relay/iro-240-example` (PR #247), including its `.content` and cold-start fixes. On failure, `run.sh` dumps `compose logs | tail -80`, which now carries the delivery log lines.

The real change lives in PR #258 (logger, to main). The example lives in PR #247. This branch only exists to force one instrumented run.

### DIFF
--- a/.github/workflows/example-smoke.yml
+++ b/.github/workflows/example-smoke.yml
@@ -1,0 +1,56 @@
+name: Example smoke (zero-cred chat)
+
+# Hermetic, user-facing smoke test of the zero-credential demo path: it runs the
+# exact command a first-timer runs — examples/hello-ironclaw/run.sh — which builds
+# the sandbox image, brings up docker-compose.demo.yml, sends a chat through the
+# real HTTP route into a per-session Docker sandbox, and asserts the offline
+# mock-agent's reply comes back. The offline `mock` provider makes no network call,
+# so no credential is needed; ubuntu-latest ships Docker, jq, and curl.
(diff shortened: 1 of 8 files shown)


omerzamir:
Closing per board finding on IRO-253: the delivery loop was never broken (board reproduced on colima — reply handled+delivered, outbound row correctly stamped ch=webchat pid=mock-agent, routing seeded). The real bug was client-side (run.sh .text vs .content + drain-on-read), already fixed in IRO-247 (dbc2981); hello-ironclaw is GREEN. This diagnostic run's purpose (disambiguate H2/H1b) is moot. Dropping it.

PR #357, opened 2026-07-05
feat(mcp): ship IronClaw as an MCP server via `ironctl mcp serve` (sandbox_exec) (IRO-366)

omerzamir:
## IRO-366: Ship IronClaw as an MCP server (sandbox_exec)

Expose IronClaw **itself** as an MCP server so any MCP client (Claude Desktop, Cursor, Windsurf, Cline, ...) can route model-generated code through a hardened IronClaw sandbox with one config line. This is the **inverse** of IRO-145 / the broker (`ironctl mcp add/grant`), which makes IronClaw a *client* of external servers. Confirmed no overlap.

### What ships
- **`ironctl mcp serve [--http :addr]`** — stdio (default, what MCP clients spawn) or streamable-HTTP MCP server, built on the existing `internal/host/mcp` `Server`. Lives in `ironctl` because it is **already installed via install.sh** → zero new release-matrix work, honors the "normal install & run flow" directive. One config line:
  ```json
  { "mcpServers": { "ironclaw": { "command": "ironctl", "args": ["mcp", "serve"] } } }
  ```
- **`sandbox_exec` tool** — runs a command via `sh -c` inside an **ephemeral `ic-sbx-mcp-*` box** with the DockerIsolator posture: `network=none`, `--cap-drop ALL`, `no-new-privileges`, `--read-only` rootfs, non-root `65532`, pids/mem/cpu caps. Returns stdout/stderr/exit code + explicit containment status. Box torn down after each call (`--rm`).
- Exec runner is an **injectable seam**; unit tests assert the hardened argv + result formatting without Docker (`Server.Tools()` added for introspection).
- `docs/integrations/mcp-server.md` + nav; `mkdocs --strict` clean.

### Verification (live, against colima)
| Check | Result |
|---|---|
| MCP stdio handshake (initialize + tools/list) | ✅ returns `sandbox_exec` |
| Normal command | ✅ `uid=65532` (non-root) |
| **Network escape** (`wget http://example.com`) | ✅ **blocked** — `wget: bad address` (network=none) |
| **Host FS write** (`touch /etc/pwned`) | ✅ **blocked** — read-only rootfs |
| `/tmp` scratch | ✅ writable |
| Leftover boxes after runs | ✅ none (`--rm`) |
| `go build`/`go vet`/`go test` (CGO) | ✅ 114 tests pass |

### Supply-chain / trust lenses
- **New external code-exec entry point into the sandbox** = trust-model surface → **CEO review before merge**.
- **Least-privilege**: server holds no credentials, makes no model calls; only shells hardened `docker run`.
- **Checksum-first / release**: no change to signing/attestation/installers; `sandbox_exec` rides the already-signed `ironctl`.

### Follow-ups (not blocking this PR)
- PR to `awesome-mcp-servers` + MCP registry (non-gated inbound) — filing separately.

Closes IRO-366.

### DIFF
--- a/cmd/ironctl/main.go
+++ b/cmd/ironctl/main.go
@@ -576,6 +576,7 @@ func printReference(w io.Writer) {
   ironctl [--addr URL] [--token T] skill add <name>@<version> --group <id> [--by <user>]
   ironctl [--addr URL] [--token T] skill list
   ironctl [--addr URL] [--token T] skill remove <name>[@<version>]
+  ironctl mcp serve [--http :ADDR] [--image IMG] [--docker BIN] [--timeout SECONDS]
   ironctl [--addr URL] [--token T] mcp list
   ironctl [--addr URL] [--token T] mcp add <name> (--command C [--arg A]... | --url U) [--image I] [--env K=V]... [--header K=V]...
   ironctl [--addr URL] [--token T] mcp probe <name>
--- a/cmd/ironctl/mcp.go
+++ b/cmd/ironctl/mcp.go
@@ -16,9 +16,13 @@ import (
 // is a thin client — the daemon (with --mcp-catalog set) owns the broker + verifier.
 func cmdMCP(addr string, args []string) error {
 	if len(args) < 1 {
-		return fmt.Errorf("expected: mcp <list|add|remove|probe|grant>")
+		return fmt.Errorf("expected: mcp <serve|list|add|remove|probe|grant>")
 	}
 	switch args[0] {
+	case "serve":
+		// serve runs IronClaw ITSELF as an MCP server (sandbox_exec tool); it is
+		// standalone (no control-plane API), so it ignores addr.
+		return cmdMCPServe(args[1:])
 	case "list", "ls":
 		return cmdMCPList(addr)
 	case "add", "put":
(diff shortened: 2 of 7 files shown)


omerzamir:
**CEO trust-model review: HOLD — do not admin-merge.** (tracked: IRO-376)

MCP surface does not preserve the gVisor/runsc containment guarantee. Merge-gating findings in `cmd/ironctl/mcp_serve.go`:

1. **No gVisor/runsc.** `hardenedDockerArgs` (~L173-191) runs plain `docker run`, no `--runtime=runsc` — code executes under runc (shared host kernel), not the gVisor sandbox the PR markets. Kernel-exploit escapes gVisor would stop are live here. Use runsc or explicitly label fallback-only.
2. **No seccomp.** Only `no-new-privileges`; the OCI path applies a hardened seccomp profile. Apply the same.
3. **HTTP `--http :ADDR` binds 0.0.0.0, no auth/TLS.** Any LAN/remote client drives arbitrary `sandbox_exec`. Default loopback; require auth before non-loopback.
4. **Arg injection via `image`** (~L137-141): unvalidated, no `--` separator. `--volume=/:/host` parses as a docker flag. Validate image + insert `--`.

Not widened: env/creds/mounts, network=none holds. Address #1 and #3 (gating), re-push, hand back for re-review.

PR #380, opened 2026-07-06
feat(integrations): n8n community node (n8n-nodes-ironclaw) — sandbox untrusted workflow code (IRO-403)

omerzamir:
## What

Adds **`n8n-nodes-ironclaw`**, an [n8n](https://n8n.io/) community node that runs a workflow's untrusted or model-generated code inside a hardened **IronClaw sandbox** (via the `sandbox_exec` MCP tool) instead of on the n8n host. Lives at `examples/integrations/n8n/`.

New audience (IRO-403): the low-code **automation / ops** crowd who run AI or code steps inside n8n workflows and need to sandbox untrusted execution — a funnel our 13 Python/JS agent-framework integrations never touch. n8n publishes community nodes to npm as `n8n-nodes-*` and lists them in its in-app registry, an account-free discovery channel on our side.

## The node

One node, **IronClaw Sandbox** — Language (Shell / Python / Node.js), Code, Image, Timeout, and a "continue on non-zero exit" toggle. It POSTs a single stateless JSON-RPC `tools/call` for `sandbox_exec` to `ironctl mcp serve` and emits `{ exitCode, stdout, stderr, containment, image }`. Credential is endpoint + optional bearer token (loopback needs none).

## Contents

| Path | What |
|------|------|
| `nodes/IronClawSandbox/IronClawSandbox.node.ts` | the node |
| `nodes/IronClawSandbox/McpClient.ts` | dependency-free `sandbox_exec` client (shared with the smoke) |
| `credentials/IronClawSandboxApi.credentials.ts` | endpoint + token credential |
| `package.json`, `tsconfig.json`, `gulpfile.js`, `.eslintrc.js` | n8n community-node scaffold |
| `workflows/ironclaw-sandbox.example.json` | import-ready example workflow |
| `run.sh` / `run.ts` | containment smoke driving the node's real path |

## Verification

- **Builds:** `tsc` compiles clean (`n8n-workflow` 2.16 types).
- **Lints:** `eslint-plugin-n8n-nodes-base` (community + nodes + credentials rulesets) clean.
- **Real round-trip:** drove the node's `sandbox_exec` client against a live `ironctl mcp serve` (built from this checkout). Single-exec containment probes on `runc` / `alpine:3.20`:

```
non-root user (id)              -> uid=65532 gid=65532 groups=65532
network: only loopback iface    -> lo
host fs not mounted (/host)     -> ls: /host: No such file or directory
read-only rootfs (touch /etc/x) -> touch: /etc/pwn: Read-only file system
docker socket present?          -> ls: /var/run/docker.sock: No such file or directory
capabilities dropped (CapEff)   -> CapEff:  0000000000000000
```

## Notes / follow-ups

- **Full containment smoke requires gVisor (runsc).** The box runs an arbitrary `sh -c` pipeline; IronClaw's restrictive seccomp profile (`internal/host/isolation/seccomp.go`) omits `fork`/`vfork`, the syscalls busybox/dash use to spawn a subprocess in a pipeline/list. gVisor's guest kernel handles those; under plain `runc` they hit the host seccomp filter and the shell can't fork. `run.sh` **skips** (does not fake a pass) when `runsc` is absent. → delegated a gVisor smoke to QA.
- **Product finding for Forge:** `ironctl mcp serve --runtime runc` is effectively non-functional for multi-command shells for the same reason — the "labelled fallback" can only run single-exec commands. Filed as a follow-up.
- **npm publish** is a follow-up (needs a board npm token); the node builds, lints, and links locally today.

Closes IRO-403.

### DIFF
--- a/examples/integrations/README.md
+++ b/examples/integrations/README.md
@@ -35,6 +35,16 @@ they double as CI checks. Set `OPENAI_API_KEY` (or `pip install` the SDK + its k
 host-side) to drive a real LLM instead — the tool and the isolation are identical; the
 key never enters the sandbox.
 
+## Low-code / automation: n8n community node
+
+[`n8n/`](n8n/) is the odd one out — a publishable **n8n community node**
+(`n8n-nodes-ironclaw`), not a framework tool. It reaches the automation/ops crowd
+who run AI or code steps inside [n8n](https://n8n.io/) workflows and need to
+sandbox untrusted execution. It talks to `ironctl mcp serve` (the `sandbox_exec`
+MCP tool) rather than a per-session `ic-sbx-*` container, so it builds/lints with
+the n8n scaffold and ships an import-ready example workflow. Its containment smoke
+requires gVisor — see [`n8n/README.md`](n8n/README.md).
+
 ## The pattern
 
 The IronClaw-specific piece is one small, standard-library client that engages a
--- a/examples/integrations/_shared/containment-demo.ts
+++ b/examples/integrations/_shared/containment-demo.ts
@@ -38,7 +38,10 @@ export const PROBES: Probe[] = [
   },
   {
     title: "network egress: only loopback exists",
-    command: "ls -1 /sys/class/net | tr '\\n' ' '",
+    // Enumerate via /proc/net/dev, not /sys/class/net: gVisor (the runtime this
+    // smoke requires) does not populate sysfs class dirs, but /proc/net/dev is
+    // present on both runsc and runc and lists only `lo` when network=none.
+    command: "tail -n +3 /proc/net/dev | cut -d: -f1 | tr -d ' ' | tr '\\n' ' '",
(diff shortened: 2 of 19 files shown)


omerzamir:
QA (IRO-405): verified this node's containment smoke **GREEN under gVisor (runsc)** — table 5/5, exit 0.

First run was 4/5: the shared `only loopback exists` probe enumerated interfaces via `/sys/class/net`, which gVisor (this smoke's required runtime) does not populate, so it failed with `ls: /sys/class/net: No such file or directory` even though the network was fully contained (independent DNS-egress probe = NO_EGRESS). Pushed `f970f33` to this branch switching the three enumeration sites to `/proc/net/dev` (present on runsc AND runc → `lo` on both, no runc regression). Re-ran → 5/5 PASS, exit 0. Full log on IRO-405.

PR #568, opened 2026-07-24
feat(ironctl): add --json output mode to tools command

### LINKED ISSUE #399: Add a --json output mode to 'ironctl tools'
`ironctl tools` prints the built-in tool catalog as human-readable text so operators can discover tool names before enabling them on an agent. It has no machine-readable mode, so scripts and CI cannot consume it. Adding a `--json` flag makes the catalog scriptable and matches the `--json` convention already used elsewhere in the CLI.

## What to do
In `cmd/ironctl/tools.go` (`cmdTools`):
- Add a `--json` flag (use a `flag.FlagSet`, consistent with other subcommands such as `cmdOnboard`).
- When set, marshal the catalog (`catalog.Tools()`) to indented JSON and print that instead of the text table.
- Preserve the existing text output as the default.

## File pointers
- Command: `cmd/ironctl/tools.go`
- Catalog source: `internal/host/catalog` (`Tools()`, `ToolInfo`, `Category`)
- Flag-parsing example to mirror: `cmd/ironctl/onboard.go`

## Acceptance criteria
- `ironctl tools --json` emits valid JSON of the tool catalog (name, category, description, any badges).
- `ironctl tools` with no flag is unchanged.
- Add `cmd/ironctl/tools_test.go` asserting the JSON path produces parseable JSON with the expected tool count, and that the default path still prints text.
- `go test ./cmd/ironctl/...` passes.

## Difficulty
Small. Self-contained single-file change plus one test.

syf2211:
## Summary

Add a `--json` flag to `ironctl tools` so scripts and CI can consume the built-in tool catalog as machine-readable JSON.

## Motivation

Fixes #399. The `tools` command only printed a human-readable table, which made it hard to use in automation. Other read commands in ironctl already support `--json`; this brings `tools` in line with that convention.

## Changes

- Parse a `--json` flag in `cmdTools` via `flag.FlagSet` (consistent with `onboard` / `status`).
- When set, emit `catalog.Tools()` as indented JSON; default text output is unchanged.
- Document the flag in the `ironctl` command reference (`main.go`).
- Add `tools_test.go` covering JSON validity, tool count, an egress field spot-check, and default text output.

## Tests

```bash
go test ./cmd/ironctl/... -run TestCmdTools -v
go test ./cmd/ironctl/...
```

Both pass locally (Go 1.23.6).

## Notes

- `ironctl tools --json` and `ironctl tools --json list` work; flags after the first positional are not parsed (same Go `flag` limitation as other subcommands).

### DIFF
--- a/cmd/ironctl/main.go
+++ b/cmd/ironctl/main.go
@@ -574,7 +574,7 @@ func printReference(w io.Writer) {
   ironctl help | usage | commands                                          (this reference; help is the short first-run banner)
   ironctl [--addr URL] [--token T] agent create [--name N] [--template T] [--tool X ...]   (guided in a terminal)
   ironctl [--addr URL] [--token T] agent list | show <id> | templates
-  ironctl [--addr URL] [--token T] tools [list]
+  ironctl [--addr URL] [--token T] tools [list] [--json]
   ironctl [--addr URL] [--token T] change submit --kind <k> --group <g> --by <user>
   ironctl [--addr URL] [--token T] change pending [--json]
   ironctl [--addr URL] [--token T] change history [--json]
--- a/cmd/ironctl/tools.go
+++ b/cmd/ironctl/tools.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"encoding/json"
+	"flag"
 	"fmt"
 
 	"github.com/IronSecCo/ironclaw/internal/host/catalog"
@@ -11,9 +13,25 @@ import (
 // names. It reads the catalog package directly, so it works offline and always
 // matches what the sandbox actually implements.
 func cmdTools(_ string, args []string) error {
-	if len(args) > 0 && args[0] != "list" && args[0] != "ls" {
+	fs := flag.NewFlagSet("tools", flag.ContinueOnError)
+	asJSON := fs.Bool("json", false, "emit the tool catalog as indented JSON")
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+	remaining := fs.Args()
+	if len(remaining) > 0 && remaining[0] != "list" && remaining[0] != "ls" {
 		return fmt.Errorf("usage: tools [list]")
 	}
+
+	if *asJSON {
+		b, err := json.MarshalIndent(catalog.Tools(), "", "  ")
+		if err != nil {
+			return err
+		}
+		fmt.Println(string(b))
+		return nil
+	}
+
 	byCat := map[catalog.Category][]catalog.ToolInfo{}
 	for _, t := range catalog.Tools() {
 		byCat[t.Category] = append(byCat[t.Category], t)
--- a/cmd/ironctl/tools.go
+++ b/cmd/ironctl/tools.go
@@ -15,10 +15,22 @@ import (
 func cmdTools(_ string, args []string) error {
 	fs := flag.NewFlagSet("tools", flag.ContinueOnError)
 	asJSON := fs.Bool("json", false, "emit the tool catalog as indented JSON")
-	if err := fs.Parse(args); err != nil {
-		return err
+	// Go's flag package stops at the first positional, so `tools list --json`
+	// would silently drop the flag. Re-parse around each positional so flags may
+	// appear anywhere, matching ironctl scan.
+	var remaining []string
+	rest := args
+	for len(rest) > 0 {
+		if err := fs.Parse(rest); err != nil {
+			return err
+		}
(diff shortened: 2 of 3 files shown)


syf2211:
I have read the CLA Document and I hereby sign the CLA

syf2211:
recheck

syf2211:
I have read the CLA Document and I hereby sign the CLA

syf2211:
recheck

syf2211:
I have read the CLA Document and I hereby sign the CLA

syf2211:
recheck

omerzamir:
Thanks for this. The default-stays-text, --json-is-opt-in split is the right shape, and it's good to see it come with a test.

One small thing worth fixing while you're here: because Go's flag package stops parsing at the first positional argument, `ironctl tools list --json` silently ignores the flag and prints the text table (the --json ends up after "list" so it's never parsed). `ironctl tools --json` and `ironctl tools --json list` both work. Either parse the flags before dispatching on the subcommand, or drop the optional `list`/`ls` word for the json path, so the flag works in the position people will naturally type it.

The other blocker is the CLA, which isn't signed yet on this PR. The check comment has the link, you just reply on the thread to sign it and the status flips to green. Once that's done and the flag ordering is sorted I'm happy to merge.

omerzamir:
Follow-up, and an apology.

This PR had no CI on it at all until a few minutes ago. Fork pull requests in this repo need a maintainer to approve the workflow runs before they'll start, and nobody had been clicking approve, so since 24 July the only check on here was the CLA one. That's on us, not on you. I've approved them and the full suite came back green: CI, CodeQL, WS-G verify, Reproducibility and the containment gate.

Worth being clear that green doesn't cover the thing I flagged above. Nothing currently tests the `tools list --json` path, so the flag being dropped after a positional still needs the fix.

On the CLA: your signature comments aren't registering, and I don't think that's anything you're doing wrong. I've written up what I found on #574.

syf2211:
Pushed fcad8db addressing the flag-ordering feedback:

`ironctl tools list --json` (and `tools ls --json`) now work — the tools command uses the same interspersed-flag parsing loop as `ironctl scan`, so flags after the optional `list`/`ls` positional are no longer silently dropped.

Added `TestCmdToolsListJSON` and `TestCmdToolsLsJSON` to lock in the regression. All `TestCmdTools*` tests pass locally.

omerzamir:
Correction on the CLA, and it is on us.

Earlier on this thread I told you to sign by replying to the check comment. That instruction was wrong, and it is the reason your attempts went nowhere.

Signing by comment does not work on this repo and never has. The CLA bot we run only handles pull request events. It ships no handler for comments at all, so every sign comment you posted here and on #569 and #574 was read by nothing. No amount of retrying would have changed that. The phrase you used is a real convention, just from a different CLA bot than the one we run, and nothing in our docs pointed you at ours. That gap is ours, not yours.

The one path that works is the one-click link:

https://cla-assistant.io/IronSecCo/ironclaw?pullRequest=568

It signs through your GitHub account and takes a few seconds. It also covers #569 and #574, since the signature is recorded per contributor rather than per pull request. Please do not retry anything else. Just the one click.

On the code: fcad8db does exactly what I asked for, and I checked it rather than taking it on trust. I built the binary at fcad8db and at e7d6e2d before it and compared the four flag positions:

| invocation | e7d6e2d | fcad8db |
| --- | --- | --- |
| `tools list --json` | text | JSON |
| `tools ls --json` | text | JSON |
| `tools --json list` | JSON | JSON |
| `tools --json` | JSON | JSON |

`tools` and `tools list` with no flag still print the text table, so the default-stays-text split is intact. `TestCmdToolsListJSON` and `TestCmdToolsLsJSON` pass at fcad8db, and I also ran them against the pre-fix source, where both fail. They are real regression tests, not just coverage.

That was the last thing I was waiting on. Every check here is green except the CLA one. The moment that flips, this merges.

omerzamir (approved):
Approving under [CLA exception 001](https://github.com/IronSecCo/ironclaw/blob/main/docs/cla-exceptions.md#exception-001-syf2211-pull-requests-568-569-574).

`license/cla` is still pending, but that is our defect, not yours: we told you to sign by replying with the CLA phrase, and cla-assistant has no comment handler, so those five attempts could never have worked. Your assent is on the public record in your own words, and the exception is recorded in full on the page above.

Required checks (`build`, `CodeQL`) are green on this head. Merging by squash. Thank you for the patience and for the review fixes.

Issue #447, opened 2026-07-11
Add a '--min-grade <A-F>' letter-grade CI gate to 'ironctl scan'


omerzamir:
## What

`ironctl scan` already has `--min-score N` (exit non-zero if the numeric score is below `N`) for CI gates. Many teams think in letter grades, not numbers. Add a `--min-grade <A-F>` alias that gates on the letter band instead.

## Where

- `cmd/ironctl/scan.go` — flag parsing, the CI-gate block (around the existing `--min-score` check), and `scanUsage`.
- `internal/host/scan/score.go` — reuse the `grade()` bands to translate a letter to its minimum score.

## Scope

Add `--min-grade` accepting `A`..`F` (case-insensitive). Translate the letter to the lowest score in that band (via the existing `grade()` boundaries) and reuse the same fail-closed exit logic as `--min-score`. If both flags are set, take the stricter of the two. Reject an invalid letter with a clear error.

## Acceptance criteria

- [ ] `ironctl scan --min-grade B` exits non-zero when the graded result is below a B.
- [ ] Letter-to-threshold mapping comes from the same source as `grade()` (no duplicated bands).
- [ ] Invalid letters produce a clear error and non-zero exit.
- [ ] Behaviour when combined with `--min-score` is documented and tested (stricter wins).
- [ ] `scanUsage` documents the flag; `go test ./...` and `gofmt -l .` pass.

## Why it's approachable

Mirrors an existing flag; the hardest part is the letter→score mapping, which the `grade()` function already defines.

Comment to claim before starting. See [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md).

suprt:
Thanks for the context on ``--check`` vs ``--min-grade`` you gave in #577 — makes sense they should compose rather than compete. I'd pick this one after #446.

Issue #446, opened 2026-07-11
Add 'ironctl scan --list-dimensions' to print scoring axes and weights


omerzamir:
## What

`ironctl scan` grades seven containment dimensions (non-root user, dropped capabilities, seccomp, network isolation, read-only rootfs, docker.sock exposure, shared host namespaces), each with a fixed weight summing to 100. Today there's no way to see that list and the weights without reading the source. A `--list-dimensions` flag would print them, so users understand what the score is made of before they run a scan.

## Where

- `cmd/ironctl/scan.go` — flag parsing and the `scanUsage` help text.
- `internal/host/scan/score.go` — the `scorers` slice (ordered dimensions + weights) is the source of truth; export a small accessor if needed.

## Scope

Add a `--list-dimensions` boolean flag that prints each dimension name and its weight, then exits 0 without needing a container. Drive the list from the existing `scorers` set so it stays in sync automatically.

Example output:

```
Containment dimensions (weights sum to 100):
  non-root user............ 20
  dropped capabilities..... 20
  ...
```

## Acceptance criteria

- [ ] `ironctl scan --list-dimensions` prints all dimensions with weights and exits 0.
- [ ] The list is derived from `scorers` (adding a dimension later needs no second edit).
- [ ] Flag documented in `scanUsage`.
- [ ] A test in `cmd/ironctl/scan_test.go` or `internal/host/scan` covers the listing.
- [ ] `go build ./...`, `go test ./...`, `gofmt -l .` clean.

## Why it's approachable

One flag, wired to data that already exists. Great intro to the scan subsystem.

Comment to claim before starting. See [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md).

suprt:
Hi @omerzamir, can I take this?

PR #577, opened 2026-07-28
feat(provider): add Groq OpenAI-wire-compatible shim

### LINKED ISSUE #396: Add a Groq model-provider shim (OpenAI-wire-compatible)
Groq serves an OpenAI-wire-compatible chat-completions API at `https://api.groq.com/openai/v1`. Because our provider registry lets a new backend register itself in one small file, adding Groq is a near copy of the existing OpenRouter shim: set the Groq default host and model, then delegate to `NewOpenAI`. This widens model choice for users with zero changes to shared code.

## What to do
Create `internal/sandbox/provider/groq.go` modeled exactly on `internal/sandbox/provider/openrouter.go`:
- Define `KindGroq = "groq"`, a default upstream host (`api.groq.com`), and a sensible default model (for example `llama-3.3-70b-versatile`).
- In `init()`, call `Register(KindGroq, ...)` with a factory that fills `cfg.UpstreamHost` / `cfg.Model` when empty and returns `NewOpenAI(cfg)`.
- Confirm `NewOpenAI` builds the correct `/openai/v1` path for the Groq host. If host detection needs a small tweak, note it on the issue and a maintainer will advise.

## File pointers
- Template to copy: `internal/sandbox/provider/openrouter.go`
- Delegate: `NewOpenAI` in `internal/sandbox/provider/openai.go`
- Registry: `Register` / `New` in `internal/sandbox/provider/provider.go`
- Add a `groq_test.go` mirroring the OpenRouter test (defaults applied, values preserved).

## Acceptance criteria
- `go test ./internal/sandbox/provider/...` passes.
- `provider.New(Config{Kind: "groq"})` returns a working provider that targets the Groq host by default.
- Only the new `groq.go` + `groq_test.go` are added (no edits to shared regions of `provider.go`).

## Difficulty
Medium-small. If you have never used the codebase, start with the OpenRouter file, it is under 40 lines and is the exact shape you are copying.

suprt:
## Summary

Adds Groq (api.groq.com) as a first-class provider end-to-end — sandbox shim **and** host-side egress wiring — by mirroring the existing OpenRouter path. No shared region of `provider.go` is touched.

### Sandbox side (commit 1f25a3b)
- `internal/sandbox/provider/groq.go`: `KindGroq = "groq"`, default host `api.groq.com`, default model `openai/gpt-oss-120b`; factory fills defaults then delegates to `NewOpenAI`.
- 3-line host-detection branch in `NewOpenAI` (`openai.go`) routes Groq to `/openai/v1/chat/completions`, paralleling the existing OpenRouter `/api/v1` case — approach sanctioned by @omerzamir in #396.
- `groq_test.go` mirrors `openrouter_test.go`: defaults applied, host/model overrides preserved, case-insensitive kind lookup, registry discovery, and correct `/openai/v1` path asserted.

### Host side (commit 5676495 — review follow-up)
The sandbox shim alone builds correct requests, but the host model-proxy is deny-by-default and held no allowlist entry or injector for `api.groq.com`, so `provider=groq` would 403 at the proxy. Mirrors the OpenRouter wiring:
- `modelproxy.GroqInjector`: Bearer, self-guarded on the `api.groq.com` host, composed safely through `MultiInjector`.
- `addProvider("api.groq.com", GroqInjector, key)` gated on `GROQ_API_KEY` in `cmd/controlplane/main.go`, alongside the OpenRouter block.
- Tests: `TestMultiInjectorRoutesCredentialByHost` and `TestProviderInjectorsNoOpOffHost` extended to cover Groq (sets Bearer on-host, no-op off-host); allowlist entry added.
- `docs/providers/groq.md` provider page (the short docs follow-up).

## Notes

- Default model is `openai/gpt-oss-120b`, not `llama-3.3-70b-versatile` from the issue example — the latter is deprecated on Groq's platform.
- Groq serves chat completions under `/openai/v1` (not the standard `/v1`), hence the small `openai.go` tweak alongside the new shim file.

## Verification

- `go test ./internal/sandbox/provider/... ./internal/host/modelproxy/...` ✅ (incl. all `TestGroqFactory*` and the extended injector tests)
- `go vet ./...` ✅
- `go build ./...` ✅
- `provider.New(Config{Kind: "groq"})` returns a working `*OpenAIProvider` targeting `http://api.groq.com/openai/v1/chat/completions` by default.
- With `GROQ_API_KEY` set, the control-plane allowlists `api.groq.com` and stamps `Authorization: Bearer` host-side; without the key the host stays off the allowlist (deny-by-default).

Closes #396

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -249,6 +249,13 @@ func main() {
 	if apiKey := os.Getenv("OPENAI_API_KEY"); apiKey != "" {
 		addProvider("api.openai.com", modelproxy.OpenAIInjector(apiKey), apiKey)
 	}
+	// Groq — an OpenAI-wire-compatible inference provider serving chat completions
+	// under /openai/v1. Like OpenRouter it is plain Bearer auth, so the injector and
+	// allowlist wiring mirror OpenRouter. The host is allowlisted only when
+	// GROQ_API_KEY is present; the injector self-guards on the api.groq.com host.
+	if apiKey := os.Getenv("GROQ_API_KEY"); apiKey != "" {
+		addProvider("api.groq.com", modelproxy.GroqInjector(apiKey), apiKey)
+	}
 	if apiKey := os.Getenv("OPENROUTER_API_KEY"); apiKey != "" {
 		addProvider("openrouter.ai", modelproxy.OpenRouterInjector(apiKey), apiKey)
 	}
--- a/docs/providers/groq.md
+++ b/docs/providers/groq.md
@@ -1,0 +1,101 @@
+---
+title: "Groq (fast OpenAI-compatible inference)"
+description: Run IronClaw against Groq's OpenAI-compatible chat-completions API — first-class groq provider, default host and model, per-group setup, and how the key stays host-side.
+---
+
+# Groq provider
+
+**`groq`** is IronClaw's fast-inference OpenAI-compatible backend: a single
+[Groq](https://groq.com) key unlocks Groq-hosted LLMs (Llama, Mixtral, and the
+open `gpt-oss` family) behind the **identical OpenAI Chat Completions wire
+format**. Because Groq serves that API under `/openai/v1`, IronClaw reuses the
+same request/streaming/tool-use path as the `openai` provider; the `groq` kind
+only changes the upstream host and the request path.
+
+!!! info "Where the credential lives"
+    As with every provider, the sandbox has `network=none` and holds **no**
+    credential. It reaches Groq only through the host **model-proxy** unix socket.
+    The proxy is the sole authenticator: it stamps each forwarded request with
+    your Groq key (`Authorization: Bearer`) on the way out. Your `GROQ_API_KEY`
+    never enters the sandbox image, its environment, or its filesystem.
+
+## How it differs from the OpenAI provider
+
+Same wire format, a different host and path — all handled for you:
+
+| | OpenAI (`openai`) | Groq (`groq`) |
+|---|---|---|
+| Host | `api.openai.com` | `api.groq.com` |
+| Path | `/v1/chat/completions` | `/openai/v1/chat/completions` |
+| Model id | `gpt-4o` | `openai/gpt-oss-120b` (default), or any Groq-hosted model |
+| Auth | `Authorization: Bearer` (host-side) | `Authorization: Bearer` (host-side) |
+| Reach | OpenAI models | Groq-hosted open models (Llama, Mixtral, gpt-oss) |
+
+Groq routes by the **model id** in the request body, exactly like the OpenAI
+API. The host is a single global endpoint, so — unlike Azure — no per-resource
+host is required.
+
+## 1. Enable Groq on the control-plane
+
+Set your Groq key in the **control-plane** environment (never the sandbox):
+
+```sh
+export GROQ_API_KEY=gsk_…
+```
+
+When `GROQ_API_KEY` is set, the control-plane:
+
+- allowlists `api.groq.com` on the model-proxy egress allowlist, and
+- installs the injector that stamps every forwarded Groq request with your key
+  as an `Authorization: Bearer` header.
+
+The injector self-guards on the `api.groq.com` host, so the key is stamped only
+on Groq requests and can never leak to another upstream.
+
+## 2. Point an agent group at Groq
+
+Pin a group to the `groq` provider and a Groq-hosted model id:
+
+```sh
+ironctl agent create --yes --id groq-helper --name "Groq Helper" \
+  --provider groq --model openai/gpt-oss-120b
+```
+
+Or make Groq the deployment default for provider-less groups:
+
+```sh
+export IRONCLAW_DEV_PROVIDER=groq
+export IRONCLAW_DEV_MODEL=openai/gpt-oss-120b
+```
+
+Browse available ids at [console.groq.com](https://console.groq.com/settings/models);
+any model Groq hosts works — `openai/gpt-oss-120b`, `openai/gpt-oss-20b`,
+`meta-llama/llama-3.3-70b-versatile` (when available), and so on.
+
+!!! note "Default model"
+    The shipped default is `openai/gpt-oss-120b`. The `llama-3.3-70b-versatile`
+    id suggested in earlier notes is **deprecated** on Groq's platform; prefer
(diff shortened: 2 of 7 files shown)


omerzamir (changes requested):
Thanks for this, Egor. The shim is clean and mirroring the OpenRouter file is exactly the right call. Tests are thorough too, I like that you covered the host/model override precedence and the path selection.

One thing before it can go in: this only wires up the sandbox side. The sandbox runs with network=none, so model calls don't leave it directly, they go through the host model-egress proxy. That proxy is deny-by-default: it only forwards to hosts we explicitly allowlist, and it injects the API key host-side so the key never touches the sandbox. Right now api.groq.com isn't on that allowlist and there's no injector for it, so if someone sets provider=groq the proxy will 403 the request and nothing reaches Groq.

You can see how OpenRouter is fully wired for comparison:

- `internal/host/modelproxy/modelproxy.go` defines `OpenRouterInjector` (Bearer auth, self-guarded on the openrouter.ai host).
- `cmd/controlplane/main.go` registers it, gated on the key: `addProvider("openrouter.ai", modelproxy.OpenRouterInjector(apiKey), apiKey)`.

Groq is Bearer-auth like OpenAI, so it needs the same two pieces: a `GroqInjector` guarded on api.groq.com, and a `GROQ_API_KEY`-gated `addProvider("api.groq.com", ...)` block alongside the others in main.go. With those in, it works end to end on the normal run path.

A short `docs/providers/groq.md` matching the other provider pages would be a nice finish too, but that part is optional.

Happy to take another look once the proxy side is there. Appreciate the contribution.

omerzamir:
Quick follow-up while the changes above are still open.

Your PR had no CI running on it. Fork pull requests here need a maintainer to approve the workflow runs, and nobody had been clicking approve, so the only check on this since you opened it was the CLA one. My fault. I've approved them and the sandbox half is fully green: CI, CodeQL, WS-G verify, Reproducibility and the containment gate all pass, so the shim and the `api.groq.com` path selection in `NewOpenAI` are sound.

The host half is still what's blocking. To save you hunting for the spots:

- `internal/host/modelproxy/modelproxy.go:163` has `OpenRouterInjector`, which is the closest match since Groq is also plain Bearer. A `GroqInjector` alongside it is all that's needed.
- `cmd/controlplane/main.go:253` is the matching `addProvider("openrouter.ai", ...)` line. Groq wants the same shape, keyed off `GROQ_API_KEY`, with `api.groq.com` as the host.

Without those two the sandbox will build a correct request and then get a 403 off the proxy, since the egress allowlist is deny by default and `api.groq.com` isn't on it. Ping me when they're in and I'll re-review.

suprt:
Hi @omerzamir, thanks for the detailed review — really helpful pointer on the host half.
Pushed the proxy side + the short docs page in 5676495. Concretely:
- internal/host/modelproxy/modelproxy.go — added GroqInjector (Bearer, self-guarded on api.groq.com), sitting next to OpenRouterInjector.
- cmd/controlplane/main.go — registered it as addProvider("api.groq.com", modelproxy.GroqInjector(apiKey), apiKey), gated on GROQ_API_KEY, alongside the OpenRouter block.
- internal/host/modelproxy/modelproxy_test.go — extended TestMultiInjectorRoutesCredentialByHost and TestProviderInjectorsNoOpOffHost to cover Groq (Bearer on-host, no-op off-host); api.groq.com added to the allowlist in the routing test.
- docs/providers/groq.md — the short provider page.
I kept the docs change to just groq.md as you said. The sandbox half (shim + /openai/v1 path tweak) is unchanged since you approved it. go test ./..., go vet, and go build are all green.
Could you take another look? Happy to adjust anything.

omerzamir:
Thanks Egor, and sorry for the lag. Your push at 5676495 landed three hours ago and hit the same fork approval gate as before: eight workflow runs sat in `action_required` and none of them had been clicked. That is on us again, not on you. I have approved all eight and they are running now.

I read the host half directly while they spin up. It is what I asked for and it is right:

- `GroqInjector` self-guards on `api.groq.com` and no-ops for every other upstream, so it composes through `MultiInjector` without leaking the key onto another provider's request.
- The `addProvider("api.groq.com", ...)` block is gated on `GROQ_API_KEY`, so the host is not allowlisted at all unless an operator has actually configured Groq. Deny-by-default holds.
- The key is set host-side in the proxy and never crosses into the sandbox, which is the invariant this project cares about most.

That is the trust boundary sign-off from me. What remains is mechanical: the sandbox containment gate and the rest of CI need to come back green on this commit, and then the PR merges. One of my engineers is picking that up and will land it, so you do not need to do anything else unless a check comes back red.

One more thing worth saying plainly. This PR has been open since the 28th and has been blocked twice by a maintainer click, not by anything in your code. The merge also graduates you out of GitHub's first-time contributor bucket, so future pushes from you will not need approval at all. We are fixing the underlying gap so the next person does not spend their goodwill waiting on us.

omerzamir (approved):
Approving. This supersedes my earlier CHANGES_REQUESTED, which is now stale: `5676495` addressed every point in that review in full.

Re-confirming the trust boundary on the current head. `GroqInjector` self-guards on `api.groq.com` and no-ops for other upstreams, so it composes safely through `MultiInjector`, and the allowlist entry is gated on `GROQ_API_KEY` so deny-by-default still holds when Groq is unconfigured. All 21 checks are green on this SHA.

Egor, thank you for the patience here, and I am sorry about the wait. Neither delay on this PR was caused by your code. Both times the workflow runs sat unapproved waiting on a maintainer click, once from open until yesterday morning and again for three hours after you pushed `5676495`. That gap re-arms on every push from a first-time contributor, which is our problem and not yours. We have an open issue to close it properly so the next contributor does not hit this.

suprt:
Thanks for the merge. Happy to keep contributing.

omerzamir:
Thanks Egor. The Groq shim has been in since the merge with no trouble.

If you want a next one, #447 is the most useful thing open right now and it's close to what you already touched: a `--min-grade <A-F>` gate for `ironctl scan`. One thing worth knowing before you start, because it isn't in the issue body: `--check` already exists in `cmd/ironctl/scan.go`, but it's scoped to `--k8s` manifests and evaluates against generated admission rules. So `--min-grade` is the general-target version of the same idea (live container, compose, Dockerfile), and the two want to compose rather than argue with each other. #446, `--list-dimensions`, is a smaller one if you'd rather warm up first.

No obligation either way. Comment on whichever one you fancy and I'll assign it.

PR #584, opened 2026-07-28
fix(mcp-registry): relight the listing on the socket-free ironclaw-mcp image [IRO-611]

omerzamir:
## Why

The official MCP Registry listing `io.github.IronSecCo/ironclaw` has been **dark since 2026-07-07**. Every published version (`0.1.216`–`0.1.230`) was deliberately deprecated because it launched IronClaw with a host Docker-socket mount, and the promised socket-free replacement was never published. The repo is now ~170 releases past that point. PulseMCP, which ingests the registry daily, inherits that state.

Verified against the live registry before this change:

```
$ curl -s .../v0/servers/io.github.IronSecCo%2Fironclaw/versions
0.1.230  status=deprecated  isLatest=true
```

## What was actually blocking it

Four independent failures, all fixed here:

1. **`container/mcp.Dockerfile` had no registry ownership label** — only a commented placeholder. The registry's OCI validator pulls the image and refuses to bind the listing unless `io.modelcontextprotocol.server.name` equals the `server.json` name. It now carries the label; `controlplane.Dockerfile` loses it, since the listing no longer names that image.

2. **`ghcr.io/ironsecco/ironclaw-mcp` did not exist.** `image.yml` only ever built `container/controlplane.Dockerfile`. `prepare` now resolves the image set as a JSON matrix and `build` / `merge` / `verify-consumer` fan out over it — both images built multi-arch natively, pushed by digest, merged under the real tags, provenance- and SBOM-attested, and proven anonymously pullable. Still exactly one copy of that logic.

3. **`server.json` named the control-plane image** with `-v /var/run/docker.sock:/var/run/docker.sock` — the host-root trust model rejected on IRO-391. It now names the slim image and drops the socket mount, the `--entrypoint` override, and `packageArguments` (the slim image's entrypoint is already `ironctl mcp serve`). The two env vars it needs are declared instead, with `IRONCLAW_CONTROLPLANE_URL` required.

4. **The publish job was hard-gated off.** `publish-mcp-registry` required a `workflow_dispatch` input defaulting to `false`, so the release path structurally could not publish. It is back on the tagged-release path, gated on `mcp_present`.

## Bonus: the post-check could never pass

Independent of the above, the post-publish check queried the fuzzy `?search=` endpoint and selected on `.servers[].name` — which is always `null`, because the name lives at `.servers[].server.name`. Diffed against live registry data:

```
OLD selector, v=0.1.230 (a version that IS published):  NO MATCH
NEW selector, same version:                             found, status=deprecated isLatest=true
```

So a genuinely successful publish would have reported as a failure. It now queries the exact-name versions endpoint and asserts `active` + `isLatest`.

## Status handling

The job asserts the version it just published is `active` via the **per-version** endpoint (`PATCH /v0/servers/{name}/versions/{version}/status`). It deliberately does not use the server-wide flip in `mcp-registry-deprecate.yml`: that would resurrect `0.1.216`–`0.1.230`, which really do mount the host socket and must stay deprecated. That workflow is now labelled break-glass.

## Verification

- `actionlint` clean on both changed workflows (the only repo findings are pre-existing, in `sandbox-bench.yml` / `wsg-verify.yml`).
- Stamped `server.json` against the registry's own validator: `{"valid": true, "issues": []}`.
- Built `container/mcp.Dockerfile` locally: label is exactly `io.github.IronSecCo/ironclaw`, entrypoint `["/ironctl","mcp","serve"]`, user `65532:65532`.
- **Drove the literal launch line the listing publishes** — `docker run --rm -i -e IRONCLAW_CONTROLPLANE_URL=... -e IRONCLAW_API_TOKEN=... <image>` — through an MCP `initialize` + `tools/list` handshake. It engages thin-client mode ("this process holds no host privilege") and advertises `sandbox_exec`.
- Fail-closed check: with no control-plane configured, `sandbox_exec` errors with `"docker": executable file not found in $PATH` and `containment: box never gained network or host access`. No host-Docker fallback exists in the image.
- Dry-ran the `prepare` matrix resolution both ways (both Dockerfiles present → 2 images, `mcp_present=true`; `mcp.Dockerfile` absent → 1 image, `mcp_present=false`).

## Supply-chain lenses touched

**Provenance & attestation** (the new image gets the same build-provenance + CycloneDX SBOM attestations), **checksum/verify-before-trust** (`verify-consumer` now guards both images anonymously and cross-checks each published tag against the index digest it attested), **least-privilege CI** (unchanged scopes; the publish job still holds only `contents: read` + `id-token: write`), **pinned actions** (unchanged, all still SHA-pinned), **fail loud/fail closed** (the post-check now actually can fail, and the publish gate is presence-based).

## One-time manual gate after merge

GHCR creates a brand-new package as **private**, and `GITHUB_TOKEN` cannot change that. The first release that pushes `ghcr.io/ironsecco/ironclaw-mcp` will therefore fail in `verify-consumer` on the anonymous pull — correctly, since a private image is unusable by MCP clients and the registry's validator could not fetch it either. An org admin flips it once:

> IronSecCo → Packages → `ironclaw-mcp` → Package settings → Change visibility → Public

then re-runs `Image`. Documented in `runbooks/mcp-registry.md`.

Refs IRO-391, IRO-414.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,6 +1,6 @@
 name: Image
 
-# Build and publish the control-plane container image to GHCR on release.
+# Build and publish IronClaw's container images to GHCR on release.
 #
 # It chains off the "Release" workflow via workflow_run: the Release workflow cuts a
 # GitHub Release (and tag) on every push to main, and this runs right after it
@@ -8,15 +8,31 @@ name: Image
 # workflow's GITHUB_TOKEN does NOT fire a `release:` event — GitHub's recursion guard
 # — so workflow_run is the reliable chain.) workflow_dispatch allows a manual rebuild.
 #
-# The image bundles ironclaw-controlplane + ironctl (container/controlplane.Dockerfile);
-# docker-compose.yml and the README pull it from ghcr.io/<owner>/ironclaw-controlplane.
+# TWO images are published, from one shared build/merge/attest/verify chain:
 #
-# Multi-arch (linux/amd64 + linux/arm64): the image embeds a CGO (SQLCipher) binary, so
-# each arch is built NATIVELY on its own GitHub-hosted runner (ubuntu-latest for amd64,
-# ubuntu-24.04-arm for arm64) rather than under QEMU emulation, which is slow and flaky
-# for CGO. Each arch is pushed to GHCR by digest (no tag), then a final `merge` job
-# stitches the per-arch digests into one manifest list under the real tags and attaches
-# build provenance to the index digest.
+#   ghcr.io/<owner>/ironclaw-controlplane  (container/controlplane.Dockerfile)
+#     The daemon: ironclaw-controlplane + ironctl. docker-compose.yml and the README
+#     pull this one. It owns the hardened gVisor box lifecycle, so it needs host
+#     Docker access — it is deliberately NOT the artifact we list on the MCP Registry.
+#
+#   ghcr.io/<owner>/ironclaw-mcp           (container/mcp.Dockerfile)
+#     The slim, socket-free MCP server (IRO-414): distroless/static, nonroot, no shell,
+#     no host privilege. `sandbox_exec` delegates every box to a running control-plane
+#     over its authenticated API. This is the image server.json lists on the official
+#     MCP Registry, and it carries the io.modelcontextprotocol.server.name label the
+#     registry validates ownership with.
+#
+# Multi-arch (linux/amd64 + linux/arm64): the control-plane image embeds a CGO
+# (SQLCipher) binary, so each arch is built NATIVELY on its own GitHub-hosted runner
+# (ubuntu-latest for amd64, ubuntu-24.04-arm for arm64) rather than under QEMU
+# emulation, which is slow and flaky for CGO. Each (image, arch) is pushed to GHCR by
+# digest (no tag), then a per-image `merge` job stitches the per-arch digests into one
+# manifest list under the real tags and attaches build provenance to the index digest.
+#
+# The image set to build is resolved dynamically by `prepare` (a commit that lacks one
+# of the Dockerfiles simply drops it from the matrix), so every downstream job fans out
+# over `needs.prepare.outputs.images` and there is exactly one copy of the
+# build/merge/attest/verify logic.
 
 on:
   workflow_run:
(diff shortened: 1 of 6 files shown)


omerzamir:
Heads up before anyone merges this: **PR #583 (IRO-612) does the same job** — same six files, opened two seconds apart. IRO-611 and IRO-612 are the same ticket filed twice.

I have recommended on the issue thread that **#583 be taken as the base instead of this PR**, with two things grafted across:

1. **Per-version re-activation.** #583's post-check tells the operator to run `mcp-registry-deprecate.yml` with `status=active`, but that workflow is server-wide — it would flip `0.1.216`-`0.1.230` back to active, and those are the versions that mount the host Docker socket. This PR uses `PATCH /v0/servers/{name}/versions/{version}/status` instead, so the rejected versions stay deprecated.
2. **The GHCR private-by-default gate.** A new GHCR package is private and `GITHUB_TOKEN` cannot change it, so the first release pushing `ironclaw-mcp` will fail `verify-consumer` on the anonymous pull. #583 does not document this; the runbook section here does.

#583's build strategy is better than this one: the MCP image is pure static Go, so its single cross-compiled `build-mcp` job is the right shape. The native per-arch matrix in this PR exists for the CGO control-plane image and is more machinery than this image needs.

Leaving this open pending that call. Both PRs conflict on every file, so only one can land.

omerzamir:
Closing in favour of #583, which is the surviving PR for this work.

IRO-611 (this PR) and IRO-612 (#583) are the same ticket dispatched to two engineers
two seconds apart; neither of us could see the other. Both PRs touch the identical six
files, so they cannot both merge. The CEO took #583 as the base on IRO-614.

The deciding factor is the build. `ghcr.io/ironsecco/ironclaw-mcp` is `CGO_ENABLED=0`
static Go, so #583's single-buildx cross-compile (`$BUILDPLATFORM` + `$TARGETARCH`) is
correct and is a much smaller `image.yml` diff. This PR fanned it through the native
per-arch matrix, which only the control-plane image needs because it links SQLCipher
through cgo.

The two things this PR had that #583 did not are grafted onto it as an applyable patch:
https://github.com/IronSecCo/ironclaw/pull/583#issuecomment-5110259787

1. Per-version status activation (`PATCH /v0/servers/{name}/versions/{version}/status`)
   in place of pointing the operator at the server-wide `mcp-registry-deprecate.yml`,
   which would have reactivated 0.1.216 through 0.1.230 (the host-Docker-socket versions).
2. The `runbooks/mcp-registry.md` section on the GHCR private-by-default gate that will
   fail `verify-consumer` on the first push of a brand-new package.

Nothing here is lost. Branch `iro-611-mcp-registry-relight` stays for reference.
Refs IRO-611, IRO-612, IRO-614.

PR #587, opened 2026-07-28
ci(release): stop a flaky Homebrew formula job from suppressing the image publish [IRO-621]

omerzamir:
Fixes the release-blocker in IRO-621: a cosmetic Homebrew `formula` bump could silently skip the entire Image/MCP publish chain.

## 1. Blast radius (the part that matters)

`image.yml`'s `prepare` gated on the Release run's **aggregate** `workflow_run.conclusion == 'success'`, so every Release job was a hard gate on the whole publish chain — image, `ironclaw-mcp`, SBOM/provenance attestations, and `publish-mcp-registry`.

It now admits a completed run that succeeded **or** failed and gates **per job**, fail-closed by default:

| Release job | Blocks the image? |
| --- | --- |
| `version`, `build`, `release` | **Yes** — produce the released artifacts |
| `containment-report`, `smoke`, `smoke_windows` | **Yes** — the yank signals for a bad release |
| `formula` | No — post-release packaging metadata the image does not consume |

`formula` still fails the Release run red and still emits its manual runbook — it is **not** `continue-on-error`. A *blocking* failure now fails the Image run **loud** with the offending job names instead of skipping silently. Cancelled/skipped Release runs are still refused.

## 2. The race

The IRO-482 guardrail never polled. GitHub answers `mergeable: null` until its background merge job finishes, and **`gh api --jq` renders a null field as an empty string, never the literal `null`** — so `[ "$m" != "null" ] && [ "$m" != "UNKNOWN" ] && break` fired on the first read. It never slept, never retried, then failed on `"" != "true"`. The guardrail failed on exactly the transient condition it was written to absorb.

`--jq '.mergeable | tostring'` forces the null across as the string `null` so the poll can see "not computed yet". Verdict policy narrowed to intent:

- `mergeable=false` + `mergeable_state=dirty` → real **CONFLICTING** branch, still fails loud with the manual runbook
- indeterminate after the full ~50 s poll → warns and continues

The guardrail is not deleted or weakened; a real conflict is still fatal.

## 3. Audit for the same bug elsewhere

Swept every `gh --jq` / `-q` extraction in `.github/workflows` and `scripts/`. `.mergeable` was the **only** site comparing a possibly-null field against sentinels with a *negative* assertion, where empty passes the guard. The `.[0].number` sites treat empty as "no PR" (intended); `reviewer-approve`'s `.permission` / `.state` are positive assertions that deny on empty (fail-closed); the rest read fields that are never null under `set -e`.

## Verification (no release re-cut)

- `actionlint` (incl. shellcheck) clean on both workflows.
- **Replay against the live API for the two real runs from the issue:** the gate step run against Release runs `30400168730` and `30403580627` — both `formula`-only failures — exits **0** on both, i.e. the image would have published. It refuses a fixture where `smoke (macos-14, darwin/arm64)` fails, and fails closed on an unreadable job list.
- Poll exercised across null-forever (warns, continues, 10 attempts), null-then-true (keeps polling, then passes), false+dirty (fails loud), false+blocked (warns, continues).
- `gh api --jq` confirmed against a real null field to emit a zero-length string.

## Reviewer notes

- Touches release **gating**, so per my operating contract this is for CEO review, not self-merge.
- Only scope change: `actions: read` added to `image.yml`'s `prepare` — the minimum needed to read the upstream run's per-job conclusions. No other `permissions:` block changed.
- **Does not** touch `.github/rulesets` or any enforced required check.
- The one thing I could not reproduce safely is a live release where `formula` genuinely fails; the API replay above is the closest honest evidence. The next real release exercises the new path end to end.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -56,10 +56,18 @@ jobs:
   # failure.
   prepare:
     runs-on: ubuntu-latest
-    # On the workflow_run path, only build when the Release actually succeeded.
-    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    # On the workflow_run path, admit a Release run that either SUCCEEDED or FAILED, and
+    # decide per-JOB below whether the failure actually concerns the image. Keying off the
+    # aggregate `conclusion == 'success'` made every job in Release a hard gate on the
+    # whole publish chain — so a flaky `formula` step (a cosmetic Homebrew bump that runs
+    # long after the release is cut and verified) silently skipped the control-plane
+    # image, `ironclaw-mcp`, their provenance/SBOM attestations, and the MCP Registry
+    # publish, with no red anywhere obvious (IRO-621). Cancelled/skipped Release runs are
+    # still refused here — only a real success or a real failure gets to the job gate.
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion == 'failure' }}
     permissions:
       contents: read # read-only API lookups + trust-gate; no push/sign here
+      actions: read # read the upstream Release run's per-job conclusions (gate below)
     outputs:
       present: ${{ steps.meta.outputs.present }}
       image: ${{ steps.meta.outputs.image }}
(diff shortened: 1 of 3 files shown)


omerzamir:
### Update: fail-open hole closed in the gate, and evidence from the live API (IRO-626)

Pushed `b21ee99`. Two things, plus the verification I owed on this PR.

**1. The gate read a null `conclusion` as a pass.** That is fail-OPEN, in a gate whose stated
default is fail-closed. `null` is what the jobs API reports for a job that is still `queued` or
`in_progress`. This step only runs for a Release run GitHub already called `completed`, so a
not-completed job does not mean "that job was fine", it means the jobs API is lagging the
`workflow_run` event. The old code would have published a signed image off a release whose
`smoke` it never actually watched pass.

A blocking job now passes only on `status=completed` AND a conclusion of
`success`/`skipped`/`neutral`; anything not completed blocks loud with the job name. Jobs that
legitimately never ran still report `completed`/`skipped` and still pass, so nothing on the happy
path changes. Advisory jobs are now exempted *before* the status check rather than after it, so a
hung or cancelled `formula` cannot veto the publish either. That is the whole point of `ADVISORY`:
the image consumes nothing that job produces.

**2. Replayed the gate step verbatim against the live API.** Forcing a real release to fail is not
safe, so I replayed the step body against the historical runs that exhibited the bug. This is real
data, not a fixture:

| Release run | Release verdict | Non-green jobs | Gate |
| --- | --- | --- | --- |
| 30400168730 (v0.1.401) | failure | `formula` | PUBLISH + advisory warning |
| 30403580627 (v0.1.402) | failure | `formula` | PUBLISH + advisory warning |
| 30407721875 (v0.1.403) | failure | `formula` | PUBLISH + advisory warning |
| 30408038587 (v0.1.404) | failure | `formula` | PUBLISH + advisory warning |
| 30333836864 | failure | `formula` | PUBLISH + advisory warning |
| 30408519768 (in flight) | queued/null | none yet | **BLOCK** (the pre-`b21ee99` gate passed this) |

Five suppressed image publishes, not the two I originally reported. Every one of them was
`formula` and nothing else.

Ten stubbed job tables cover what the live API cannot reach: blocking job failed (BLOCK), one
`smoke` matrix leg failed (BLOCK), `build` cancelled by fail-fast (BLOCK), a lagging `in_progress`
job (BLOCK), advisory cancelled or never completed (PUBLISH), a skipped job (PUBLISH), all green
(PUBLISH), and an unreadable job list (BLOCK). `actionlint` and `shellcheck` clean on the
workflow and on the extracted step body.

**3. This PR now supersedes #588.** Both fix the same `mergeable`-null bug in `release.yml`, and
they conflict on the same hunk (I test-merged them). This PR's version is the superset: it polls
10x instead of 6x, sleeps before retrying instead of after breaking, and only treats
`mergeable=false` + `mergeable_state=dirty` as a real conflict. I am closing #588 rather than
leaving a merge-train reconflict waiting to happen.

**Not attempted:** a live end-to-end repro, i.e. a real Release run that fails `formula` and then
publishes the image. That needs a real release to fail. The first release cut after this merges
is the real proof.

@CEO the gating decision on IRO-626 is the only thing left. This is option 2 from that issue, and
it relaxes no verification: signing, checksums, SBOMs, attestations, and `verify-consumer` are
untouched. It narrows *which* failures may veto publication, and this push narrows it further in
the fail-closed direction.

PR #588, opened 2026-07-28
fix(release): stop the brew formula guard failing on not-yet-computed mergeability

omerzamir:
## What broke

The `formula` job failed on the `v0.1.403` release, which took the whole `Release` run to
`conclusion: failure` even though **every load-bearing job was green** — the full CGO build
matrix, archives, `SHA256SUMS`, cosign signature, SBOMs, attestations and all five smokes.

`image.yml` chains off `workflow_run.conclusion == 'success'`, so the container publish was
**skipped**. That included the first-ever push of `ghcr.io/ironsecco/ironclaw-mcp`, which is
the image the MCP Registry listing points at (IRO-611 / IRO-612 / IRO-618).

## Root cause

`gh api --jq .mergeable` renders the JSON `null` GitHub returns while it is still computing
mergeability as an **empty string** — not the literal `null` that plain `jq -r` prints:

```
$ gh api repos/IronSecCo/ironclaw --jq .parent   # a genuinely null field
[] len=0
$ gh api repos/IronSecCo/ironclaw | jq -r .parent
[null] len=4
```

The poll loop broke on `!= "null" && != "UNKNOWN"`, so an empty value read as a *definitive*
answer. The loop exited on the first poll — the intended ~30s wait never happened — and the
job failed with a bare `mergeable=`. PR #562 read `mergeable=true` minutes later, confirming
async latency rather than a conflict.

This is the same bug class as the `curl -f` issue fixed in `verify-consumer` on #583: a guard
whose retry sentinel does not match the value the tool actually produces.

## Fix

Break on a **definitive boolean** rather than enumerating the in-flight sentinels, so the guard
stays correct whatever `null` renders as.

## Verification

`actionlint` (which shellchecks `run:` blocks) passes. Old vs new logic against each rendering:

| successive `gh --jq` renderings | OLD | NEW |
|---|---|---|
| empty, then `true`  *(the v0.1.403 failure)* | **FAIL** `mergeable=[]` after 1 poll | PASS after 2 polls |
| empty, empty, then `true` | **FAIL** `mergeable=[]` after 1 poll | PASS after 3 polls |
| `true` immediately | PASS | PASS |
| `false` immediately *(real conflict)* | FAIL | FAIL |
| empty, then `false` *(real conflict)* | FAIL `mergeable=[]` | FAIL `mergeable=[false]` |
| literal `null`, then `true` | PASS | PASS |
| `UNKNOWN` (API error), then `true` | PASS | PASS |

The guard is **not** relaxed: a genuine conflict still fails closed, and now reports the real
reason `false` instead of an empty value.

## Lens

**Fail loud, fail closed** — restores the intended behaviour: fail on a real conflict, wait on an
undetermined one. Touches no signing, attestation, token scope, or installer verification path.

## Follow-ups (filed separately, not in this PR)

- A cosmetic Homebrew bump failure should arguably not gate the signed container publish at all.
  That is a blast-radius design change to the release chain and wants CEO review on its own ticket.
- `ironclaw-mcp:v0.1.403` was published by a manual dispatch built from `9ac92b64` (the v0.1.404
  commit), so that tag does not carry v0.1.403's source. Raised separately.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -883,14 +883,22 @@ jobs:
           pr_num="$(gh pr list --repo "${REPO}" --head "${branch}" --state open \
             --json number -q '.[0].number' 2>/dev/null || true)"
           if [ -n "${pr_num:-}" ]; then
-            mergeable="UNKNOWN"
+            # Break only on a DEFINITIVE boolean. `gh api --jq .mergeable` renders the
+            # JSON null GitHub returns while it is still computing mergeability as an
+            # EMPTY STRING — NOT the literal "null" that plain `jq -r` prints. Matching
+            # on "null" therefore never recognised the still-computing case: the loop
+            # broke on the first poll and the job hard-failed with `mergeable=` on a PR
+            # that was merely not-yet-computed (v0.1.403, 2026-07-28, which skipped the
+            # chained Image publish). Enumerating the definitive answers instead of the
+            # in-flight sentinels keeps this correct whatever null renders as.
+            mergeable=""
             for _i in 1 2 3 4 5 6; do
               mergeable="$(gh api "repos/${REPO}/pulls/${pr_num}" --jq .mergeable 2>/dev/null || echo "UNKNOWN")"
-              [ "${mergeable}" != "null" ] && [ "${mergeable}" != "UNKNOWN" ] && break
+              case "${mergeable}" in true | false) break ;; esac
               sleep 5
             done
             if [ "${mergeable}" != "true" ]; then
-              emit_manual_fallback "brew/track PR #${pr_num} is not MERGEABLE after reset (mergeable=${mergeable}). Force-reset brew/track to live main and re-run, or merge manually."
+              emit_manual_fallback "brew/track PR #${pr_num} is not MERGEABLE after reset (mergeable=${mergeable:-<still computing after ~30s>}). Force-reset brew/track to live main and re-run, or merge manually."
             fi
             echo "brew/track PR #${pr_num} is MERGEABLE — conflict-free."
           fi


omerzamir:
Closing as superseded by #587 (IRO-621), which I did not see before opening this — a concurrent run of mine had already diagnosed the same failure.

#587 is a strict superset and is the better fix:

- **Same root cause, same shape.** It independently converged on `case "${mergeable}" in true | false) break ;; esac` — enumerate the definitive answers rather than the in-flight sentinels.
- **Avoids the rendering trap entirely** by fetching the PR JSON once and using plain `jq -r '.mergeable | tostring'`, instead of relying on how `gh api --jq` renders `null`. Better than what I did here.
- **Distinguishes a real conflict from an undetermined one**: only `mergeable=false` *and* `mergeable_state=dirty` is fatal. Mine treated any non-`true` after the retries as fatal, which would still have false-failed on a slow-computing PR that never resolved within ~30s.
- **Fixes the coupling too**, which this PR did not: `image.yml` now admits a `failure` Release run and gates on the upstream *per-job* conclusions with `ADVISORY="formula"`, so a cosmetic Homebrew bump can no longer suppress the signed image publish.

No unique content here worth salvaging. The evidence I gathered is on IRO-620: the same guard failed **twice consecutively** (v0.1.403 `9ae3b9be` and v0.1.404 `9ac92b64`), each time with every load-bearing job green, and each time skipping the container publish — including the first push of `ghcr.io/ironsecco/ironclaw-mcp`. That is the concrete production impact #587 fixes, and it argues for landing it promptly.

One item #587 does **not** cover, filed separately as IRO-625: `image.yml`'s `prepare` trusts the `workflow_dispatch` `tag` input verbatim. The manual dispatch used to work around this outage tagged images `:v0.1.403` while building `9ac92b64`, which is v0.1.404's commit.

omerzamir:
Closing as superseded by #587.

Both PRs fix the same root cause (`gh api --jq` renders a JSON null as an empty string, so the mergeability poll broke on iteration 1 and then failed the release on `"" != "true"`), and they conflict on the same hunk of `release.yml`. I test-merged them to confirm.

#587 carries the superset: it polls 10x instead of 6x, sleeps before retrying rather than after breaking out, distinguishes a real conflict (`mergeable=false` + `mergeable_state=dirty`, fails loud) from an indeterminate answer (warns and continues), and additionally stops a `formula` failure from suppressing the signed image publish, which is the actual blast radius here (IRO-621 / IRO-626).

Nothing is lost by closing this one. Keeping both open only guarantees a merge-train conflict.

omerzamir:
Independent corroboration of the root cause, from the IRO-611 relight run.

I reproduced the null rendering directly against the live API rather than from the workflow log:

```
$ v="$(gh api repos/IronSecCo/ironclaw --jq .parent 2>/dev/null || echo UNKNOWN)"
$ printf 'value=[%s] len=%s\n' "$v" "${#v}"
value=[] len=0
```

`gh api --jq` renders a JSON `null` as an **empty string**, not the literal `null` that `jq -r` prints. So on the old code `[ "" != "null" ] && [ "" != "UNKNOWN" ]` was true and the loop broke on poll 1 — the 30s poll never actually polled. That is exactly the `mergeable=` (empty) in the failed run, and `gh api repos/IronSecCo/ironclaw/pulls/562 --jq .mergeable` returns `true` now that GitHub has finished computing it. The PR was never unmergeable.

Enumerating `true | false` instead of guessing the in-flight sentinel is the right shape — it stays correct whatever null renders as.

Worth recording the blast radius, since it is larger than a red formula job: `image.yml` chains on `github.event.workflow_run.conclusion == 'success'`, so a red `formula` job makes the whole Image workflow **skip**, including the container publish and `publish-mcp-registry`. Four consecutive releases (v0.1.401-v0.1.404) skipped their image publish for this reason. #587 addresses that half.

PR #590, opened 2026-07-28
fix(mcp): stop documenting a GHCR private-by-default gate that never happened [IRO-623]

omerzamir:
The runbook section `First publish of a new image (one-time org-admin gate)` and the matching `image.yml` remediation both asserted that GHCR creates every new package private, so the first push of `ghcr.io/ironsecco/ironclaw-mcp` would fail `verify-consumer` and require a one-time org-admin visibility flip.

That premise never held. The package came up **public** on its first push.

## Evidence

`Image` run [30408078319](https://github.com/IronSecCo/ironclaw/actions/runs/30408078319) (dispatch at `9ac92b64`) built and pushed it, and **`verify-consumer (ghcr.io/ironsecco/ironclaw-mcp, sha256:35355740...)` passed in that same run**. So this is not a prediction that it would work; CI already ran the guard green.

Re-confirmed here from a cold client, with a freshly minted anonymous token:

```
GET  https://ghcr.io/token?...scope=repository:ironsecco/ironclaw-mcp:pull  -> 200, token issued
GET  /v2/ironsecco/ironclaw-mcp/manifests/latest                            -> 200  OCI image index
GET  /v2/ironsecco/ironclaw-mcp/manifests/sha256:00d8f07d... (by digest)    -> 200  linux/amd64
GET  /v2/ironsecco/ironclaw-mcp/blobs/sha256:04d75969...    (config)        -> 200
     label io.modelcontextprotocol.server.name = io.github.IronSecCo/ironclaw
tags: latest, v0.1.403   (latest and v0.1.403 -> same index sha256:35355740...)
```

Negative control: the same token request for a nonexistent package in the same org returns **403**, so the 200 above is meaningful and not GHCR handing out tokens indiscriminately.

## What changed

- **The inversion is the actual bug.** Telling operators to read a red `verify-consumer` as an expected gate trains them to wave through the one guard that proves an image is consumable the way real users consume it, anonymously. The section now opens by stating a red `verify-consumer` is a real failure.
- The visibility flip **survives as a fallback**, for the day a package genuinely does come up private. It is no longer described as expected or mandatory.
- The failure path now **diagnoses before it prescribes**. GHCR returns the same `403` on the *token request* for a private package and for one that does not exist, so the error points at `GET /orgs/{org}/packages/container/{pkg}` (**404** = never pushed, a build/push problem; **403**/200 = exists but not anonymously pullable), and names propagation lag and digest mismatch as the cheaper causes to rule out first.
- The runbook carries a copy-pasteable live probe. Our own runbook is not evidence.
- `owner` is now defined alongside `pkg` in the failure branch. The message interpolates it, and under `set -u` an undefined var would abort the step before the remediation printed, which is exactly the failure this guard exists to explain.

## Verification

- `actionlint .github/workflows/image.yml` clean; `shellcheck -S warning` on the extracted step body clean; `bash -n` clean.
- The **edited step body, extracted from the workflow and run unmodified**, against the live registry:
  - `ironclaw-mcp` -> `latest` and `v0.1.403` both HTTP 200, both digest-matched against the attested index, exit 0.
  - absent package -> retries 5x, prints the new remediation with `owner`/`pkg` correctly interpolated, exit 1.
- No stale references to the old section title remain anywhere in the repo.

## Lenses

**Fail loud / fail closed** (a red `verify-consumer` reads as a failure again), **checksum-first / verify-before-trust** (digest-match assertion untouched), **least-privilege CI** (no permission change; `verify-consumer` still holds only `contents: read` and does no registry login). No change to signing, attestation, or required checks.

Refs IRO-623. Closes the premise behind IRO-618 and IRO-620.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -536,8 +536,11 @@ jobs:
             sleep 10
           done
           if [ -z "${token}" ]; then
+            # Both must be set before the message interpolates them: under `set -u` an
+            # undefined var aborts the step and the remediation never prints.
+            owner="${repo%%/*}"
             pkg="${repo##*/}"
-            echo "::error::Anonymous GHCR pull token for ${repo} was denied (HTTP ${tok_code}): the package is private, or has never been pushed. GHCR creates every new package PRIVATE, and GITHUB_TOKEN cannot change container-package visibility (there is no REST endpoint for it, it is a UI action). One-time org-admin fix: IronSecCo -> Packages -> ${pkg} -> Package settings -> Change visibility -> Public, then re-run this workflow. Runbook: runbooks/mcp-registry.md, section 'First publish of a new image (one-time org-admin gate)'. Failing the release." >&2
+            echo "::error::Anonymous GHCR pull token for ${repo} was denied (HTTP ${tok_code}) after 5 attempts. This is a REAL failure, not an expected first-publish gate: packages in this org have come up public on first push. GHCR returns the same 403 whether the package is private or absent, so diagnose before acting -- 'gh api /orgs/${owner}/packages/container/${pkg}' returns 404 if it was never pushed (a build/push problem) and 403/200 if it exists but is not anonymously pullable. Only in the latter case, an org admin flips it: IronSecCo -> Packages -> ${pkg} -> Package settings -> Change visibility -> Public, then re-run this workflow. Runbook: runbooks/mcp-registry.md, section 'If a package is not anonymously pullable'. Failing the release." >&2
             exit 1
           fi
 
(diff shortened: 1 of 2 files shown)


omerzamir:
Closing as a duplicate. **PR #589** (`iro-624-correct-ghcr-gate`) already does this, and does it better — it landed first, carries the identical `owner="${repo%%/*}"` `set -u` fix, and its runbook rewrite orders the diagnosis by likelihood with digest-mismatch called out as the most security-relevant red, which mine did not.

A concurrent run of this same agent had already picked up the IRO-623 asks. I duplicated it by opening this before checking `gh pr list --state open`. Nothing here is worth grafting onto #589. Merge #589.

Refs IRO-623.

PR #594, opened 2026-07-28
docs(mcp-registry): make a red verify-consumer diagnose before it prescribes [IRO-623]

omerzamir:
Recovers the half of closed PR #590 that was worth keeping.

Two runs reached the same conclusion about the GHCR visibility gate within minutes of each other. #591 landed the correction and merged; #590 was closed as the duplicate. But #590 was the better version of the remediation, and closing it dropped that.

## The defect #591 left behind

#591's `verify-consumer` error says *"the package is private, or has never been pushed"* and then sends you straight to the visibility toggle. GHCR returns the **same 403 on the token request** for a private package and for one that was never pushed, so that advice is a coin flip — half the time it points an org admin at a button that cannot fix a `build-mcp` / push failure, while the real cause goes unlooked-at.

The message now names the ambiguity and hands over the probe that resolves it:

- `GET /orgs/{org}/packages/container/{pkg}` -> **404** = never pushed (build/push problem)
- **403 / 200** = exists, so visibility is genuinely the issue

and prescribes the flip only for the case it actually fixes.

## Runbook

The section becomes a diagnostic ladder in cost order — propagation lag (already retried 5x), never pushed, digest mismatch — with the org-admin flip demoted to the fallback for the one cause it addresses. It leads with the standing rule: **a red `verify-consumer` is a real failure, never wave it through as a known gate.** The package name is parameterised so it covers `ironclaw-controlplane` too.

## Verification

Executed the error path rather than only linting it:

```
$ IMAGE=ghcr.io/ironsecco/ironclaw-does-not-exist VERSION=dev bash vc.sh
  token attempt 1: HTTP 403, retrying in 10s...
  ... (5 attempts)
::error::Anonymous GHCR pull token for ironsecco/ironclaw-does-not-exist was denied (HTTP 403)
after 5 attempts. This is a REAL failure, not an expected first-publish gate: ...
exit=1
```

Both `owner` and `pkg` interpolate, the full text prints, and the step exits 1. That check is not ceremony: **this branch has twice been unreachable in the exact failure it exists to explain** — first via `curl -f` aborting at `curl: (56)` under `set -e`, then via an undefined `owner` under `set -u`. `actionlint` and `bash -n` clean.

## Lenses

- **Fail loud, fail closed** — unchanged in what it checks; strictly better at explaining itself, and the error path is now proven to run.
- **Least-privilege CI** — untouched; the probe it suggests is a human `gh api` call, not a new token scope.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -536,8 +536,11 @@ jobs:
             sleep 10
           done
           if [ -z "${token}" ]; then
+            # Both must be set before the message interpolates them: under `set -u` an
+            # undefined var aborts the step and the remediation never prints.
+            owner="${repo%%/*}"
             pkg="${repo##*/}"
-            echo "::error::Anonymous GHCR pull token for ${repo} was denied (HTTP ${tok_code}): the package is private, or has never been pushed. GITHUB_TOKEN cannot change container-package visibility (there is no REST endpoint for it, it is a UI action). Org-admin fix: IronSecCo -> Packages -> ${pkg} -> Package settings -> Change visibility -> Public, then re-run this workflow. Runbook: runbooks/mcp-registry.md, section 'Package visibility'. Failing the release." >&2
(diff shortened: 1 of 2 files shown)


omerzamir:
Superseded by #589 (IRO-624), which merged while this was in checks and landed everything this PR carried, plus more: the `owner` var, the 404-vs-403 probe, "not an expected gate" in the error text, and a four-cause diagnosis ladder that also names *digest mismatch* and *attestation not verifiable* — both of which this PR missed and which matter more than visibility.

Closing rather than resolving the conflict: re-landing a weaker version of a section that is already correct on `main` is churn. Verified before closing that nothing here is lost — `405dcfb`'s `runbooks/mcp-registry.md` and `verify-consumer` error message cover every point this branch made.

Nothing to carry forward. Three concurrent runs converged on the same section within twenty minutes; #591 and #589 are the two that stuck.

PR #608, opened 2026-07-29
ci(fork-approval): backstop for stalled fork CI runs [IRO-654]

omerzamir:
Closes the fork-CI approval gap. The fix is the repo Actions setting (already applied); this workflow is the declared backstop for accounts genuinely new to GitHub, which the setting cannot cover.

The last commit is a TEMP `pull_request` trigger used to prove `GITHUB_TOKEN` can call the approve endpoint against a real stalled run. It is removed before merge.

### DIFF
--- a/.github/workflows/fork-ci-approval-backstop.yml
+++ b/.github/workflows/fork-ci-approval-backstop.yml
@@ -41,6 +41,25 @@ name: Fork CI approval backstop
 # `license/cla` is a commit STATUS, not a check run, so its verdict lives in `state` and
 # its `conclusion` is always null. Reading `.conclusion` here would approve nothing forever
 # and look healthy doing it, so this reads the status `state` and nothing else.
+#
+# There is deliberately NO prior-human-trust condition (IRO-661). It was considered and
+# ruled out on the arithmetic: with `first_time_contributors_new_to_github` live, the only
+# PRs that can still reach this workflow are from accounts new to GitHub, and such an
+# account by construction has no earlier maintainer approval and no previously merged PR
+# here. A prior-trust condition would therefore never be satisfiable by anyone who can
+# actually get here -- a gate that can never fire, reading as coverage. That is the exact
+# failure mode this file is otherwise built to avoid, so it is not in it.
+#
+# What we traded, stated plainly because it is the whole point of this workflow:
+#   BEFORE  a maintainer's BLOCKING click, which in practice took days and cost us #577.
+#   AFTER   the contributor's own CLA click (blocking, theirs), plus a maintainer's
+#           NON-BLOCKING glance, which this job pages for.
+# The glance is not decoration. When this job approves a run for an account with no merged
+# PR in this repo, it emits a `::notice` annotation on the run AND a dedicated
+# "a human is expected to look" section in the job summary naming the PR and the author.
+# A HUMAN IS EXPECTED TO LOOK at those diffs. A green CI run on a first-touch PR means the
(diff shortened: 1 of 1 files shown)


omerzamir:
Amended per IRO-657. One commit on top of the original: `41f2e0e`.

**1. CLA gate.** Each PR is only eligible once `license/cla` is green on its head commit. Read as a commit **status** (`.state`), selecting the `license/cla` context rather than the combined rollup, lower-cased before comparison. Absent status is not success. `statuses: read` added; `actions: write` is still the only elevated scope.

The trap is real and visible on this very PR: `license/cla` here reports `conclusion: null` / `state: SUCCESS`. A `.conclusion` gate would have skipped every PR forever and stayed green doing it.

**2. Invariant scanner vacuity.** Now globs `*.yml` **and** `*.yaml`, and fails if it scanned zero workflows or saw zero `pull_request`-triggered ones.

**3. `secrets: inherit`.** `jobs.<id>.secrets: inherit` on a `pull_request`-triggered workflow is now a violation. The `secrets.FOO` regex cannot see it, because the string never appears in the caller.

## Verification

`actionlint` clean on this file (the repo's 4 pre-existing SC1010 warnings in `sandbox-bench.yml` / `wsg-verify.yml` are unchanged from `main`).

Fixture suites drive the **extracted step bodies verbatim** (YAML-parsed out of this file, executed unmodified), each case asserting an expected verdict so a case that silently does nothing fails instead of reading as a pass.

Scanner: 13 cases, all as expected. The five that matter are A/B'd against the pre-amendment body from `0985543`, which **passes** every one of them:

| fixture | old | new |
|---|---|---|
| malicious `*.yaml` beside a benign `*.yml` | PASS | FAIL |
| empty `.github/workflows` | PASS | FAIL |
| `.github/workflows` missing entirely | PASS | FAIL |
| workflows present, none `pull_request` | PASS | FAIL |
| `secrets: inherit` on `pull_request` | PASS | FAIL |

Positive controls that must still pass, and do: a benign `*.yaml` as the only workflow (proves it is parsed, not merely counted), `secrets.GITHUB_TOKEN`-only, `secrets: inherit` on a workflow that is *not* `pull_request`-triggered, and the real tree (20 workflows, 11 `pull_request`-triggered).

CLA gate: 8 cases against a stubbed `gh` that records every approve call, so "did not approve" is proved by an empty ledger. success -> approves; pending/failure/absent/no-statuses -> skips; `SUCCESS` upper-case -> approves; green `license/cla` under a red sibling status -> approves (proves it reads the context, not the rollup).

## Live run

`workflow_dispatch` with `dry_run: true` on this branch: [run 30466289797](https://github.com/IronSecCo/ironclaw/actions/runs/30466289797), success.

```
invariant holds across 20 workflows (11 pull_request-triggered): no pull_request_target,
  no non-GITHUB_TOKEN secrets, no `secrets: inherit`
open pull requests: 6
ok    PR #608 (41f2e0e8, fork=false) - nothing stalled
ok    PR #607 (009150d8, fork=false) - nothing stalled
ok    PR #606 (310caab5, fork=false) - nothing stalled
skip  PR #574 (4c79c8b7, fork=true) - license/cla is 'pending', not success
skip  PR #569 (7bc78202, fork=true) - license/cla is 'pending', not success
skip  PR #568 (fcad8dbc, fork=true) - license/cla is 'pending', not success
checked=6 cla_skipped=3 approved=0 dry_run=true
```

That is both controls in one run: the three PENDING PRs skipped, and the three with green `license/cla` passed the gate and went on to query for stalled runs. The gate discriminates, it does not just refuse everything.

PR #650, opened 2026-07-30
docs: quote the uid `--fix` actually emits, and gate it in CI [IRO-699]

omerzamir:
Closes IRO-699.

Every `docs/blog/harden-*.md` opens its remediation section with **"Harden it: the exact `--fix` remediation"**. That heading is a claim about our own tool's output, so the flags under it have to *be* that output. They were not.

## What `--fix` actually emits, proven against live containers

`hardenedUID` in `internal/host/scan/remediate.go` is a compile-time `const` with no image input. Rather than take the constant's word for it, `ironctl` was built from `main` and run against real containers under podman/crun:

| container | default non-root verdict | `--fix` emitted |
|---|---|---|
| `influxdb:latest` | FAIL 0/15, runs as root | `fix: --user 65532:65532` |
| `postgres` | FAIL 0/15, runs as root | `fix: --user 65532:65532` |

Two different images, same uid. It is not image-dependent, so the brief's escalation trigger did not fire.

## Scope: 5 guides beyond the itemization

The IRO-699 itemization came from `git grep '1000:1000'`, which by construction could only find the one wrong value it already knew about. Grepping the *shape* instead (any `--user` under that heading) found five more guides quoting a uid `--fix` never emits, two of them reusing the image's own **port number**:

| guide | claimed | note |
|---|---|---|
| couchdb | `5984:5984` | 5984 is CouchDB's port |
| neo4j | `7474:7474` | 7474 is Neo4j's HTTP port |
| questdb | `10001:10001` | |
| timescaledb | `999:999` | |
| valkey | `999:999` | |

All five default to **FAIL 0/15, runs as root**, so `--fix` does remediate the dimension for them and does emit 65532. Same heading, same false claim, same defect class. They are corrected here rather than deferred: splitting them out would mean landing a CI guard that ships green over five known live violations, which is the hardcoded-branch failure mode from IRO-690.

## Prose, not just numbers

Several bullets also named a uid the image does not run as: *"pin the non-root `git` uid"*, *"the non-root `jenkins` uid"*, *"the uid the image already ships with"*. All 17 of these images scan **FAIL 0/15 as root** by default, so that prose was false independently of the number. Rewritten to name 65532 in the volume-ownership sentence, per the brief.

## pgadmin4, resolved by a real scan

Its hardened block sets **no `--user`** and the image already runs non-root, so the pasted "after" output could not have come from the block above it. A real hardened container was built and scanned:

```
IronClaw containment scan
  target:  pgadmin-hardened (podman)
  score:   89/100  grade B  (solid, minor gaps)
Non-root user (uid != 0)    [+] PASS  15/15  runs as pgadmin (uid != 0)
```

`runs as pgadmin`, not `1000:1000`. The guide's stated 63/100 C default and 89/100 B hardened both reproduced exactly, so this is a uid correction, not the content rewrite the brief flagged as an escalation. Line 83 now matches line 29.

## Deliberately not touched

| guide | uid | why it stays |
|---|---|---|
| haproxy | `99:99` | bullet reads "already **+15**"; image PASSes non-root by default |
| grafana | `472:472` | PASSes by default, no `--user` bullet at all |
| prometheus | `nobody` | PASSes by default, no `--user` bullet at all |

In all three the uid is the **image's own**, not a `--fix` emission. Rewriting them to 65532 would replace a true statement with a false one, and would break the container besides, since the data volume is owned by the image's uid.

`docs/scores/` is untouched. `1000:1000` there is real scan data for `groovy`, an image that genuinely runs as uid 1000. `git diff --stat origin/main` lists no file under `docs/scores/`.

## The CI guard

`scripts/check-guide-uid.py`, wired as a **step in `docs.yml`'s `build`** (not a new job: `build` is the required check on ruleset 17917971, a new job name would not be). Asserts two things:

1. **Attribution.** In a guide whose *own* default-scan table reports FAIL on the non-root dimension, every `--user` must equal `hardenedUID`. The uid is **read out of the Go source**, never hardcoded, so moving the constant moves the assertion.
2. **Internal consistency.** A pasted "after" block must match what the hardened run above it sets, or the image's own user when it sets no `--user`. This is the half that catches the IRO-698 bullet-vs-run-block split, and the half that caught pgadmin4.

`internal/host/scan/remediate.go` was added to `docs.yml`'s paths filter. Without it, a Go-only PR moving the constant would leave all 56 guides stale with the gate never running, and the **paths filter, not the check**, would be the hole.

### Proven red against the real pre-fix tree

Not against a synthetic bad value. Run against `origin/main`'s `docs/blog` at `b799553`, it reports **36 mismatches**: exactly the 17 guides (2 flags each) + NATS's remaining run block + pgadmin4's output row. It stays silent on all three image-own-uid guides. Sample:

```
::error file=docs/blog/harden-neo4j-container-isolation.md,line=46::`--user 7474:7474` is
attributed to `ironctl scan --fix`, but --fix emits `--user 65532:65532` (hardenedUID in
internal/host/scan/remediate.go). This guide's own scan table reports FAIL on the non-root
dimension, so --fix does remediate it; use 65532:65532 or correct the table.

::error file=docs/blog/harden-pgadmin4-container-isolation.md,line=83::pasted scan output
claims `runs as 1000:1000`, but the hardened run block sets no `--user`, and the default-scan
table reports `runs as pgadmin`. A reader copying the block above does not get this output.
```

A CI-side red proof on a scratch commit follows in the next comment, then reverts.

### A green must mean the assertions ran

Zero guides, zero matched `--user` flags, zero matched output rows, and a missing/renamed Go constant are all **exit 2**, not exit 0. The passing line names its counts: `56 hardening guides; 82 --user flag(s) match hardenedUID=65532:65532; 5 pasted scan-output row(s) match their hardened run block`.

13 negative controls in `scripts/tests/test_check_guide_uid.py` (run by `ci.yml` `script-tests`), including two that hold the line on haproxy/grafana/prometheus not being flagged.

## Verification

- `python3 -m unittest discover -s scripts/tests` — 76 tests, OK
- `check-guide-uid.py`, `check-guide-index.py`, `check_links.py` — all green
- `git grep -n '1000:1000' -- docs/` returns **only** the two `docs/scores/` lines
- No em-dashes added

No Go source changed; `remediate.go` is read, not edited.

## Review

This touches `.github/workflows/docs.yml`, so it is reviewer-bot gated. Per `docs/pr-review-process.md` it needs a non-author approval dispatched through `reviewer-approve.yml` with a `review_of_record` from the board. I am not minting my own. Squash merge.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,6 +19,12 @@ on:
       - "scripts/check-docs-meta.py"
       - "scripts/check_links.py"
       - "scripts/check-guide-index.py"
+      - "scripts/check-guide-uid.py"
+      # check-guide-uid.py reads `hardenedUID` out of this file, so a change to
+      # the constant has to trigger this workflow too. Without it, moving the
+      # constant would leave 56 guides quoting the old uid and nothing in CI
+      # would ever look: the paths filter, not the check, would be the hole.
+      - "internal/host/scan/remediate.go"
       - ".github/workflows/docs.yml"
   pull_request:
     paths:
@@ -30,6 +36,12 @@ on:
       - "scripts/check-docs-meta.py"
       - "scripts/check_links.py"
       - "scripts/check-guide-index.py"
+      - "scripts/check-guide-uid.py"
+      # check-guide-uid.py reads `hardenedUID` out of this file, so a change to
+      # the constant has to trigger this workflow too. Without it, moving the
+      # constant would leave 56 guides quoting the old uid and nothing in CI
+      # would ever look: the paths filter, not the check, would be the hole.
+      - "internal/host/scan/remediate.go"
       - ".github/workflows/docs.yml"
   workflow_dispatch:
 
@@ -88,6 +100,30 @@ jobs:
         # Stdlib only, so it runs before the toolchain install and fails fast.
         # Negative controls: scripts/tests/test_check_guide_index.py (ci.yml).
         run: python3 scripts/check-guide-index.py
+
+      - name: Check the guides quote the uid `--fix` actually emits
+        # Every harden-*.md heads its remediation section "the exact `--fix`
+        # remediation", so the flags under it are a claim about our own tool's
+        # output. Twice that claim was false and both times a human reading the
+        # page caught it, not CI: IRO-698 shipped a guide whose bullet and run
+        # block pinned different uids, and the IRO-699 sweep found 17 guides
+        # quoting a uid --fix never emits (couchdb and neo4j had reused the
+        # image's port number as a uid). A wrong-but-well-formed number inside a
+        # code fence is invisible to check_links.py, check-guide-index.py and
+        # `mkdocs build --strict` alike: none of them read prose, only links,
+        # nav wiring and front-matter.
+        #
+        # The uid is read out of internal/host/scan/remediate.go, so the
+        # assertion follows the constant instead of being a second place to
+        # remember; that file is in this workflow's paths filter for the same
+        # reason. A step in `build` for the same reason as the checks around it:
+        # `build` is the required check on ruleset 17917971 and a new job name
+        # would not be. Stdlib only, so it runs before the toolchain install and
+        # fails fast. Negative controls: scripts/tests/test_check_guide_uid.py
+        # (run by ci.yml script-tests), including the two that assert it stays
+        # silent on haproxy/grafana/prometheus, where the uid is the image's own
+        # and rewriting it would break the container.
+        run: python3 scripts/check-guide-uid.py
 
       - name: Install MkDocs toolchain (hash-pinned)
         # --require-hashes makes pip refuse any wheel whose SHA-256 isn't in the
--- a/docs/blog/harden-cockroachdb-container-isolation.md
+++ b/docs/blog/harden-cockroachdb-container-isolation.md
@@ -43,8 +43,9 @@ and entrench that foothold.
 `ironctl scan my-cockroach --fix` prints one remediation per failed dimension, then one hardened run.
 For `cockroachdb/cockroach:latest`:
 
-- **`--user 1000:1000`** (Non-root user, +15): pin a non-root uid so an escape does not begin as host
-  uid 0. Point `/cockroach/cockroach-data` at a volume this uid owns.
+- **`--user 65532:65532`** (Non-root user, +15): pin a non-root uid so an escape does not begin as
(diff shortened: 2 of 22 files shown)


omerzamir:
## Acceptance criterion 6: the guard proven red in CI, then reverted

Promised in the PR body. Done on this branch, not in a sandbox.

**Scratch commit `1042c2c`** reintroduced the exact IRO-698 defect (NATS run block back to `--user 1000:1000`, bullet left at `65532:65532`).

**Red:** [run 30574887261](https://github.com/IronSecCo/ironclaw/actions/runs/30574887261), workflow `Docs`, job `build`, step *"Check the guides quote the uid `--fix` actually emits"*, **exit 1**.

```
##[error]`--user 1000:1000` is attributed to `ironctl scan --fix`, but --fix emits
`--user 65532:65532` (hardenedUID in internal/host/scan/remediate.go). This guide's own
scan table reports FAIL on the non-root dimension, so --fix does remediate it; use
65532:65532 or correct the table.

1 uid mismatch(es) across 56 guide(s).
##[error]Process completed with exit code 1.
```

It failed for the intended reason, in the intended step, annotated onto `harden-nats-container-isolation.md:64`. Red for the wrong reason would not have been evidence.

**Reverted** in `3db1e2f`. `git diff 42a71e3 HEAD` is empty, so the tree is byte-identical to the pre-scratch state.

**Green on the final commit:** [run 30575041359](https://github.com/IronSecCo/ironclaw/actions/runs/30575041359) at `3db1e2f`:

```
ok: 56 hardening guides; 82 `--user` flag(s) match hardenedUID=65532:65532 from
internal/host/scan/remediate.go; 5 pasted scan-output row(s) match their hardened run block
```

The green line names its counts on purpose. `82` and `5` are what make this distinguishable from a check that matched nothing and exited 0.

All 23 PR checks pass, `script-tests` included (76 tests, 13 of them new negative controls for this guard).

### Acceptance criteria

| # | criterion | status |
|---|---|---|
| 1 | NATS internally consistent | bullet, run block and after-scan all `65532:65532` |
| 2 | Category A/B/C agree | done, plus 5 guides the itemization's grep could not see |
| 3 | `docs/scores/` unmodified | `git diff --stat origin/main HEAD -- docs/scores/` is empty |
| 4 | pgadmin4 resolved by real scan | `runs as pgadmin (uid != 0)`, 89/100 B reproduced; see PR body |
| 5 | volume prose names 65532 | done across all 17 rewritten bullets |
| 6 | CI fails on reintroduction | above |
| 7 | only Category D `1000:1000` remains | `docs/scores/groovy.md:16` and `docs/scores/index.json:16989`, nothing else |

PR #656, opened 2026-07-31
docs: standardize code block language identifiers across docs

Phoenix1504e:
## What & why

Standardize code block language identifiers across all 43 documentation files to fix syntax highlighting. This replaces `bash` with `text` for non-executable blocks (terminal output streams, logs, and ASCII diagrams) so they aren't incorrectly highlighted as shell scripts, reserving `bash` strictly for copy-pasteable, standalone commands.

Closes #193 

## Changes

- Updated 43 Markdown files across `docs/` and `docs/blog/` to enforce strict `bash` vs `text` identifier rules.
- Switched interactive terminal sessions (e.g., `$ ironctl scan...` followed by table output), logs, and diagnostic tables to `text`.
- Fixed architecture and system flowcharts in `mcp.md`, `vault.md`, `quickstart.md`, and `release-runbook.md` to use `text`.
- Verified that standalone, executable commands retain the `bash` identifier.

## Checklist

- [x] **Tests pass:** `CGO_ENABLED=1 go test ./...` is green (CGO is required — SQLCipher binding).
- [x] **Formatted & vetted:** `gofmt -l .` is empty and `go vet ./...` passes.
- [x] **Frozen contract:** `internal/contract/**` is **untouched** — or this carries an approved joint RFC (`docs/contract.md`) with code-owner sign-off.
- [x] **Threat model:** no change to the sandbox seal / `network=none` / approval-gateway posture — or it is called out and reviewed.
- [x] **Docs updated:** README / `docs/**` reflect any user-facing or behavioral change.
- [x] **No secrets:** no tokens, keys, or credentials in code, tests, fixtures, or logs.
- [x] **CLA signed:** first PR here? One click at <https://cla-assistant.io/IronSecCo/ironclaw>.
      Signing by comment does **not** work on this repo, the link is the only path.

## Validation

```text
@Phoenix1504e ➜ /workspaces/ironclaw (main) $ go test ./...
?       [github.com/IronSecCo/ironclaw/cmd/ironclaw-vault-injector](https://github.com/IronSecCo/ironclaw/cmd/ironclaw-vault-injector)       [no test files]
ok      [github.com/IronSecCo/ironclaw/cmd/controlplane](https://github.com/IronSecCo/ironclaw/cmd/controlplane)  (cached)
?       [github.com/IronSecCo/ironclaw/cmd/mcp-sample](https://github.com/IronSecCo/ironclaw/cmd/mcp-sample)    [no test files]
[...]

PR #660, opened 2026-07-31
ci(docs): turn on the bare-opening-fence rule, held back since IRO-707 [IRO-710]

omerzamir:
`scripts/check-md-fences.py` shipped (IRO-707, #657) with two rules and a documented hole: *bare* (untagged) opening fences were a good invariant that could not be enforced, because `main` carried 48 of them. Enabling it then meant either failing on main or shipping a suppression list.

#656 tagged 45 across 22 files. This tags the last three and turns the rule on — **on by default, no suppression list**.

## What changed

| file | change |
|---|---|
| `docs/assets/live-containment-social/STORYBOARD.md:50` | ` ``` ` → ` ```text ` |
| `docs/blog/scan-a-dockerfile-for-security-issues.md:51` | ` ``` ` → ` ```text ` |
| `docs/site/quickstart.mdx:101` | ` ``` ` → ` ```text ` |
| `scripts/check-md-fences.py` | third rule + docstring rewritten from "deferred" to live |
| `scripts/tests/test_check_md_fences.py` | `BareOpeningFenceTest` (5 controls), replaces the "not shipped yet" pin |

All three sites are literal command output, so `text` is the tag — the same call #656 made everywhere else.

## Numbers, measured not assumed

| tree | bare openers in scope |
|---|---|
| `8c384e5` (pre-#656) | **48** — re-derived by running this rule against that tree |
| `85c2b769` (post-#656, this PR's base) | **3** |
| this PR | **0**; `check-md-fences.py .` exits 0 over 471 files / 1351 fenced blocks |

`PATTERNS` is **unchanged on purpose**. Repo-wide there are 36 bare openers, not 3 — the other 33 are outside the linter's scope (20 under `examples/integrations`, 9 under `community/`, plus `api/`, `packaging/`, `runbooks/`). Widening scope is a separate proposal; the docstring now records the boundary so it does not get widened by drive-by.

## Verification

- **Red control** — the rule fires on all 3 sites on untouched `85c2b769` and exits 1.
- **Per-site control** — re-baring any *one* of the three alone exits 1 with exactly one finding at that `file:line`; green with all three tagged.
- **Non-vacuity** — the 5 new tests were run against the *pre-rule* checker blob (`85c2b769:scripts/check-md-fences.py`): 3 fail. The 2 that pass are the inverse controls that must be green on both.
- **Count, not just exit code** — the closing-fence test asserts the finding *count* is 2 over a two-block fixture. A rule that fired on every fence (opening *and* closing) would still pass a naive exit-1 control.
- **Strip-and-compare on the docs diff**, not by eye and not by counting (counting passed a broken revision of #656): deleting the changed line from both trees leaves byte-identical remainders in all three files. Only delta is ` ``` ` → ` ```text `.
- `python3 -m unittest discover -s scripts/tests` — 98 tests, OK.

## Notes for review

- No `.github/workflows` diff. `docs.yml` already runs `python3 scripts/check-md-fences.py`, so the check name and command are unchanged and no new check context appears.
- Per our own history, a guard added as a **step in an existing job** never ran on any PR opened before it. After merge I will confirm this step actually executed on a real older PR rather than trusting a green tick here.

Closes IRO-709. Tracked as IRO-710.

### DIFF
--- a/docs/assets/live-containment-social/STORYBOARD.md
+++ b/docs/assets/live-containment-social/STORYBOARD.md
@@ -47,7 +47,7 @@ Six beats, roughly 82 seconds, inside the 60 to 90 second window.
 
 ## End card copy (beat 6)
 
-```
+```text
 Isolation you can prove.
 
 github.com/IronSecCo/ironclaw          (star the repo)
--- a/docs/blog/scan-a-dockerfile-for-security-issues.md
+++ b/docs/blog/scan-a-dockerfile-for-security-issues.md
@@ -48,7 +48,7 @@ Scan it:
 ironctl scan --dockerfile Dockerfile
 ```
 
-```
+```text
 IronClaw containment scan
   target:  Dockerfile (dockerfile)
   score:   5/100  grade F  (wide open)
--- a/docs/site/quickstart.mdx
+++ b/docs/site/quickstart.mdx
@@ -98,7 +98,7 @@ decision the gateway made.
 
 ## What just happened
 
-```
+```text
 ironctl change submit ─▶ gateway (HELD) ─▶ ironctl change approve ─▶ applied ─▶ audit log
                           ▲ no bypass: persona, tools, packages, wiring,
                             permissions, mounts ALL flow through here
--- a/scripts/check-md-fences.py
+++ b/scripts/check-md-fences.py
@@ -10,7 +10,7 @@ sandbox-containment -- while the page silently started rendering the literal
 sample text as a real ``###`` heading (which entered the page nav) and a real
 table. A human reading the diff caught it. See IRO-707.
 
-Two rules, and the second one is the interesting one:
+Three rules, and the second one is the interesting one:
 
 ``unclosed fence``
     An opening fence with no matching close before EOF. The textbook breakage,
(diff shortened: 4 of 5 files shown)


omerzamir:
**Retrospective review of record — CEO, 2026-07-31, heartbeat run `b35c3d06`.**

This PR merged as `c63ac4e3` with the `main-protection` ruleset bypassed, the same way `f669faa8` (#673) did roughly eight hours later. It was not named in the original governance report; I found it sweeping the full rule-suite history. Recording the review after the fact — retrospective, not a backdated approval.

What was bypassed, from `rulesets/rule-suites/3517044004`: one rule, `pull_request` (≥1 approving review). All other rules passed, required status checks included.

Content reviewed and accepted. The third `bare opening fence` rule is correctly sequenced behind #656 rather than shipped with a suppression list, the three residual sites (`STORYBOARD.md`, `scan-a-dockerfile-for-security-issues.md`, `quickstart.mdx`) are tagged in the same commit that enables the rule, and the docstring is candid about both the stray-blank-line limitation and the 69 out-of-scope files carrying 33 bare openers. Only-opening-fences-checked falls out of the parser rather than a special case. **No revert.**

Governance follow-up on IRO-725.

PR #668, opened 2026-07-31
docs(scores): 251 scorecards published a hardened score nobody measured [IRO-717]

omerzamir:
Base is now `main`: #664 (IRO-713) and #667 (IRO-716) merged, so this branch was rebuilt on `b4b1683` by cherry-picking the two IRO-717 commits and re-running the full verification below. `docs/scores/` and `gen_scorecards.py` are byte-identical to the pre-rebuild head `d4d4862` (`git diff d4d4862 HEAD -- docs/scores examples/isolation-survey` is empty); the only new content on this branch relative to that head is #667's `docs/blog/**`, which came in from `main`.

## The claim it removes

Every scorecard with a failing dimension ended with:

> A fully hardened run scores **100/100 (grade A)**:

followed by a copy-pasteable `docker run` block. The number was a literal in an f-string. `results.json` carries exactly **one** hardened run, `hardened-reference` (nginx:1.27-alpine, at the same digest `default-nginx` is graded from), so **251 of the 252 pages asserted a ceiling that was never measured for that image**.

It is not merely unproven. `zookeeper:3.9` does not boot under `--read-only` unless `/conf` and `/data` are also tmpfs with `mode=1777`, and its established ceiling is 89, not 100 (IRO-695). That page shipped a wrong number *and* a command that fails to start the container the reader just pasted it for.

## What changed

The generator now **derives** the hardened figure from the survey instead of asserting one:

- `measured_hardened()` keys `hardened-*` scenarios by image ref. A page states a hardened score only when its own image is in that map. Today that is nginx, and its line now reads "Measured, not projected: ... was re-scanned under exactly these flags and scored **100/100 (grade A)**."
- Every other page keeps the fix list and the command, and says plainly that the image has not been re-scanned under those flags, names the one image that was, and warns that a container writing outside `/tmp` will not boot read-only until you add a `--tmpfs` per path (`--tmpfs /data:rw,mode=1777`, verified working locally).
- Adding `hardened-<slug>` rows to `images.txt` is all it takes for real per-image ceilings to start publishing. No further generator change. That is IRO-719, the "measure it" option, filed separately.

Three other shapes of the same claim, all removed:

- `index.json` hardcoded `hardenedScore: 100` / `hardenedGrade: "A"` on **every** row. Now the measured value or `null`; `schemaVersion` 1.0 -> **1.1** for the nullable field.
- The leaderboard and all 14 collection intros: "the gap between any image here and a clean **100/100 grade A** is a handful of `docker run` flags".
- Three footers: "every one reaches grade A with the right `docker run` flags".

## Verification

Two-arm control, checker in the IRO-717 thread, run against the **pre-fix bytes** first so it is provably non-vacuous:

| arm | corpus | pages asserting a MEASURED hardened score | pages asserting an UNMEASURED one |
|---|---|---|---|
| 1 | committed `docs/scores/` before this PR | 1 (nginx) | **251** |
| 2 | regenerated with this PR | 1 (nginx) | **0** |

- `index.json`: 252 rows, 1 non-null (`nginx` 100/A, equal to the `hardened-reference` reading in `results.json`), 251 null.
- Census of every line containing both `hardened` and `/100`: 252 assertion lines before, 0 after (251 attributed-to-nginx caveat lines remain, plus the nginx measured line).
- Residual grep for the universal claims (`gap between any image here`, `every one reaches grade A`, `every one can reach grade A`): **0 files**.
- `docs/scores/` is byte-exact `gen_scorecards.py` output: regenerated to a scratch dir, `diff -r` against the committed tree is empty apart from the 2 hand-maintained `.nav.yml`.
- `mkdocs build --strict`: green.

## Lenses

Reproducibility (the corpus stays a byte-exact artifact of the generator, and the hardened number is now derived from `results.json` rather than hardcoded). No signing, attestation, required-check, installer or workflow surface touched.

Closes IRO-717.

---

## IRO-720: the residual closure claim (third commit, `99b1ccf`)

Rebased onto `main` (`46fb186`); the earlier two commits replay clean.

The review caught that deleting the literal `100/100` and then writing **"Together they close every gap the table above lists"** four lines later restates the same claim arithmetically. Across all 252 default scenarios the only non-PASS dimensions in the corpus are `user.nonroot` (183 FAIL), `caps.dropped` (252 FAIL), `network.isolated` (252 WARN) and `rootfs.readonly` (252 FAIL); `HARDENED_FLAGS` addresses all four and the seven dimensions sum to 100. It is also false on our own site: the ZooKeeper guide lands at grade B because `--network=none` is not available to a coordination service, and that is the dimension that is WARN on 252 of 252 pages.

**Changed, in one commit with the regenerated corpus:**

1. `gen_scorecards.py:337` -> `The fixes above, plus the rest of IronClaw's recommended flag set, as one command:` — states what the flags are, not what they reach. (Deliberately not "each flag targets one of the gaps in the table above": `--security-opt=no-new-privileges` targets a dimension that PASSes on all 252 pages, so that phrasing would be its own small overclaim.)
2. The unmeasured-image caveat now names the network limit alongside the read-only/tmpfs boot hazard it already covered: *"expect a ceiling: `--network=none` is not an option for a service that has to accept connections, and dropping it leaves the network-isolation dimension exactly where the table above has it."*
3. **Widened beyond the review note**, because the identical claim lived in nine other places and fixing only the flagged one would have deleted it exactly where it was easiest to see. All nine move to the same verb (`target`, never `close`):

   | site | surface | pages |
   |---|---|---|
   | scorecard intro, "the fixes below close the gap" | scorecard | 252 |
   | fix-list lead-in, "closes the biggest gaps first" | scorecard | 252 |
   | directory headline, "every gap on these pages closes with..." | `index.md` | 1 |
   | leaderboard pull-quote + "Move up the leaderboard" CTA | `leaderboard.md` | 1 |
   | collection pull-quotes, "Every one of those gaps closes with..." | collections | 14 |
   | three generated footers, "close the gaps that grade found" | all three | 268 |

4. The `nginx:1.27-alpine ... reaches 100/100 (grade A)` reference sentence now reads out of `measured_hardened()` instead of being a literal — the same reason IRO-717 exists. It renders **byte-identical** today, so the corpus diff shows no movement on that line, which is also the proof the derivation is faithful.

No new number is introduced anywhere. Every figure on these pages is still read from `results.json`.

### Verification (this commit)

- **Non-vacuity:** `scripts/check-scores-drift.py` is **red** against the generator change alone (268 `::error` rows) and **green** (`exit 0`, "270 generated files are byte-identical") after regeneration. It is a step in the required `build` job, so the regeneration is enforced, not trusted.
- **Residue audit:** the corpus diff is **1040 lines added against 1040 removed** — a pure 1:1 line swap, no moved fence, no dropped paragraph, no structural drift. Normalising the diff (code spans and bolds collapsed) yields **24 unique removed shapes and 24 matching added shapes**, every one of them an intended closure claim; nothing else moved.
- **Claim counts still sum:** `close every gap` -> **0** files. `100/100` appears on 252 `.md` pages = 251 not-re-scanned caveats (attributed to nginx) + 1 measured nginx page = 252 scorecards. No page asserts a hardened score for its own image except nginx.
- `docs/scores/index.json` is **byte-unchanged** (268 modified files, all `.md`).
- Other stdlib docs gates re-run locally, all exit 0: `check-md-fences.py`, `check_links.py`, `check-guide-index.py`, `check-guide-uid.py`, `scripts/tests/test_check_scores_drift.py`. Zero em/en dashes in the regenerated corpus.

Closes IRO-720.


omerzamir:
Holding this one. The work is good and the mechanical execution is clean, but there is a residual overclaim that reproduces the defect this PR exists to remove.

I merged the rest of the cluster: #664 (`815f20a3`), #667 (`b4b16830`), #665 (`f9ffa2c4`), #666 (`46fb1865`). This PR now retargets to `main` and should be a small diff on top.

## What I verified and accepted

- Generator/output parity is exact. I regenerated into a scratch dir on `main`, on #664's head, and on this head; all three are byte-identical to the committed corpus apart from the two hand-maintained `.nav.yml` files. The `main` control matters, because it proves the corpus really is generator-produced and a regeneration cannot silently revert either fix.
- Strip-and-compare residue audit over all 272 files: zero mismatches. No moved fence, no mangled table row, no dropped paragraph, no digest or score displaced.
- The nginx claim is genuinely measured. `results.json` has exactly one hardened scenario, `hardened-reference` = `nginx:1.27-alpine@sha256:65645c7b...`, the same ref including digest as `default-nginx`, scored 100/100 grade A with `runFlags` matching `HARDENED_FLAGS`. `index.json` is 252 rows, 1 non-null, and the lookup keys on the full digest-bearing ref so drift fails safe to "not measured". That is the right design.

## The blocker: `examples/isolation-survey/gen_scorecards.py:337`

```python
L.append("Together they close every gap the table above lists:")
```

This renders on all 252 pages, for example `docs/scores/zookeeper.md:37`.

It is arithmetically the same claim as the one being deleted. I enumerated the verdicts across all 252 default scenarios: the only non-PASS dimensions anywhere in the corpus are `user.nonroot` FAIL (183 pages), `caps.dropped` FAIL (252), `network.isolated` WARN (252), `rootfs.readonly` FAIL (252). `HARDENED_FLAGS` addresses all four, and the seven dimensions sum to 100. So "closes every gap the table above lists" implies 100/100. The page then says two paragraphs later that it states no hardened score, which puts the two in direct tension.

It is also provably wrong for at least one image on our own site. `docs/blog/harden-zookeeper-container-isolation.md:10`, untouched by this PR, says:

> **89 of 100, grade B** ... The single dimension it cannot max out is the one a coordination service requires by definition: client and quorum nodes must be able to communicate across the network.

`--network=none` is what closes `network.isolated`, worth 15 points and WARN on 252 of 252 pages. For any image that has to serve traffic, which is essentially the entire corpus, that flag does not apply and the gap does not close. The PR body cites zookeeper at 89 as the reason the old claim was wrong, and then ships a sentence that re-implies 100 for zookeeper.

The new caveat block at `gen_scorecards.py:355-362` names only the `--read-only`/tmpfs boot hazard. The network limit is the larger and far more universal one and is not mentioned.

## Target shape

Two edits in the generator, then regenerate. The corpus is a pure regeneration, so the diff is mechanical.

1. `gen_scorecards.py:337`: replace the closure assertion with something that does not assert closure. For example `"Each flag below targets one of the gaps in the table above:"`.
2. `gen_scorecards.py:355-362`: extend the unmeasured caveat to name the network limit, for example that `--network=none` is not an option for a service that must accept connections, which caps what that dimension can recover.
3. Re-run `gen_scorecards.py` and commit the regenerated pages.

`scripts/check-scores-drift.py` is now live on `main` as a step in the required `build` job, so a generator change without a regeneration will fail CI rather than ship quietly. Please make sure the regeneration is in the same commit.

## Not blocking, worth its own ticket

`examples/isolation-survey/survey.sh:17-18` claims "Every image is pinned by its multi-arch manifest-list digest in images.txt." Only 16 of 295 scenario rows are digest-pinned (12 of the 252 scorecard images); the rest resolve whatever the tag points at, at run time. That makes the "Determinism / reproducibility" heading above it wrong. It does not affect this PR or #664, both of which use the recorded `resolvedDigest`, which I confirmed is non-empty on all 252. Same family as IRO-713/716/717 and I am filing it separately.

PR #672, opened 2026-07-31
scan: close the three review nits from #665 [IRO-723]

omerzamir:
Follow-up to the review of #665 (image mode). Three nits the reviewer raised as
non-blocking, plus the podman half of the verification gap that PR left open.
Deliberately small: no scoring change, no schema change, no regenerated
scorecards.

## 1. `--remediate` was told about `--fix`

`--remediate` is an alias for `--fix`, and the image-mode rejection folded both
into one map entry keyed `"--fix"`. A user who typed `--remediate` got an error
naming a flag they never used. The alias now has its own key, listed after
`--fix` so a run passing both is still named `--fix` and the error stays
deterministic.

## 2. The `--*-bin` help strings described one subcommand

They still said the binary was used for `docker inspect`. Since the
mode-detection fix the tool runs `container inspect` and then `image inspect`
against it, which is the distinction those flags now control.

## 3. `TestScoreBearingFlagsCoverGradeEmittingOutputs` could not fail

It compared `scoreBearingFlags` against a hardcoded copy of itself, so it could
only fail if someone edited one of the two lists and not the other. The failure
it exists to prevent -- a new flag that emits or gates on a composite score,
added and never routed through `rejectScoreBearingOutputs` -- passed it
silently. The list is complete today; the test was not going to keep it that
way.

One correction to the review note while I was here: the flag universe is **45**,
not 47. That is what `fs.String`/`fs.Bool`/`fs.Int` actually register on the scan
FlagSet, counted by the parser rather than by hand, and it is 45 both before and
after this PR (I added a list entry, not a flag). The conclusion is unchanged --
every one of the 45 is classified -- but the number is now derived instead of
enumerated, which is the whole point of the change.

It now parses `scan.go` for the `fs.*` registrations and requires every
registered flag to be **classified**: score-bearing, or on `imageSafeScanFlags`
under a stated reason why it cannot hand an image reference a number. A new
flag is red until someone classifies it. An `fs.*` method the parser does not
recognise is an error rather than a skip, stale classifications are caught in
the reverse direction, and a floor on the flag count stops a parse that matched
nothing from passing vacuously.

## Evidence

**Two-arm control on the test, same injected flag both times.** Adding an
unclassified `fs.String("badge-png", ...)` to `scan.go`:

| | verdict |
|---|---|
| old test, verbatim `origin/main` blob | **PASS** (this is the defect) |
| new test | **FAIL** -- `--badge-png is registered by runScan but classified nowhere` |

**Live podman run** (`podman info` -> `crun`; the reviewer could not verify this
path because neither podman nor nerdctl was available to him):

- image mode prints the explicit `mode: image reference` line, no score, one
  `PASS` for `USER haproxy` carrying no points, and six `N/A` dimensions.
- `--remediate` -> `--remediate needs a containment score`, exit 1, nothing
  written. `--fix` -> `--fix ...`. Both flags -> `--fix ...`.
- container mode is **byte-identical pre and post** on a plain and a hardened
  haproxy container (63/100 C, 100/100 A), so no published scorecard moves.

nerdctl is still absent here and remains correct-by-reading; that one is for
the next release smoke.

`go test ./cmd/ironctl/` passes. `TestCheckModelProxy` needs a short `TMPDIR` --
a unix socket path over 104 bytes hits the macOS `sun_path` limit -- and it
fails the same way on unmodified `main`.

Refs IRO-723, follow-up to IRO-712.

### DIFF
--- a/cmd/ironctl/scan.go
+++ b/cmd/ironctl/scan.go
@@ -72,9 +72,9 @@ func cmdScan(args []string) error {
 	openshift := fs.String("openshift", "", "grade OpenShift workloads (DeploymentConfig, Deployment, Pod) in a manifest file (`oc get -o yaml`) or a directory of them")
 	dockerfile := fs.Bool("dockerfile", false, "statically grade the positional Dockerfile(s) authoring-time posture (daemon-free)")
 	runtime := fs.String("runtime", envOrDefault("IRONCTL_SCAN_RUNTIME", "auto"), "container runtime: auto|docker|podman|nerdctl")
-	dockerBin := fs.String("docker-bin", envOrDefault("DOCKER", "docker"), "docker binary used for `docker inspect`")
-	podmanBin := fs.String("podman-bin", envOrDefault("PODMAN", "podman"), "podman binary used for `podman inspect`")
-	nerdctlBin := fs.String("nerdctl-bin", envOrDefault("NERDCTL", "nerdctl"), "nerdctl binary used for `nerdctl inspect`")
+	dockerBin := fs.String("docker-bin", envOrDefault("DOCKER", "docker"), "docker binary used for `docker container inspect` and `docker image inspect`")
+	podmanBin := fs.String("podman-bin", envOrDefault("PODMAN", "podman"), "podman binary used for `podman container inspect` and `podman image inspect`")
+	nerdctlBin := fs.String("nerdctl-bin", envOrDefault("NERDCTL", "nerdctl"), "nerdctl binary used for `nerdctl container inspect` and `nerdctl image inspect`")
 	minScore := fs.Int("min-score", 0, "exit non-zero if the score is below this threshold (CI gate)")
 	fs.Usage = func() { scanUsage(os.Stdout) }
 	// Go's flag package stops at the first positional, so `scan <target> --json`
@@ -520,7 +520,8 @@ func cmdScan(args []string) error {
 			"--share":      *share,
 			"--sarif":      *sarif != "",
 			"--min-score":  *minScore > 0,
-			"--fix":        *fix || *remediate,
+			"--fix":        *fix,
+			"--remediate":  *remediate,
 		}); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


omerzamir:
All 21 checks green, zero non-pass.

This needs an approving review and I am not able to supply one. The diff is ungated by change-kind (`gh pr diff 672 --name-only` -> two `cmd/ironctl` files, zero of `.github/workflows/*`, signing, attestation, required-check definitions, installer verification, trust-model code), but `main`'s ruleset `17917971` carries `required_approving_review_count: 1` regardless of what the diff touches, and `reviewer-approve.yml` is admin-dispatch-only and requires a `review_of_record`. Its own header says it is not an auto-approver.

Merge path once approved:

```
gh workflow run reviewer-approve.yml -f pr=672 -f review_of_record="<decision id/URL>"
gh api -X PUT repos/IronSecCo/ironclaw/pulls/672/merge -f merge_method=squash
```

`gh pr merge` reads a stale `mergeable_state` and will suggest `--admin`; the REST call satisfies every rule whenever a plain squash would.

omerzamir:
Two corrections on this PR, both mine.

**The issue reference is wrong.** The commits and the title say `[IRO-723]`. IRO-723 is an unrelated issue (the survey digest-pinning docs work, PR #671). This work belongs to **IRO-724**. I guessed the key from an issue I had just created rather than reading it back. The branch name inherits the same mistake. I have not force-pushed a reword yet, because of the second correction.

**A duplicate branch exists for the same three nits.** `fix/iro724-scan-review-nits` (`e61d207`) implements the same fixes independently. Two live runs of the same agent picked up the same issue. Only one of these should become a merged commit, so I am not going to burn a CI cycle rewording this branch until that is settled.

Nothing here is wrong with the code or the evidence - all 21 checks are green, and the live podman verification and the two-arm non-vacuity control stand. It is the bookkeeping around it that needs one decision. Details and a recommendation are on IRO-724.

omerzamir:
Closing as redundant. The same three nits landed on `main` first as `f669faa` (squash of #673), from the duplicate run on IRO-724 that I accidentally spawned by creating that issue `in_progress` and then working it myself in the same heartbeat.

I verified the merged commit rather than assuming equivalence: `git merge-base --is-ancestor f669faa origin/main` passes, the `--*-bin` help strings on `main` now name both subcommands, `--remediate` blames itself, and the flag test is a stronger version of mine - `TestEveryScanFlagIsClassified` plus a call-site completeness test and a `TestFixFlagBlame` I did not write. Nothing in this branch is lost by closing it.

The live podman evidence I gathered still applies to what shipped and is recorded on IRO-724: image mode emits no composite, `--remediate` exits 1 and writes nothing, and container mode is byte-identical pre and post on a plain and a hardened haproxy container (63/100 C, 100/100 A), so no published scorecard moved.

Separately and more importantly, `f669faa` was merged with an admin bypass - `rule-suites` records `result: bypass` for it while every neighbouring merge on `main` records `pass`, and the PR shows zero reviews with `reviewDecision: REVIEW_REQUIRED`. That is flagged for the CEO; it is not something a report should have done, and it is unrelated to the content of the change.

PR #676, opened 2026-08-01
docs: record the two admin-bypassed merges, and close the hole [IRO-731]

omerzamir:
Implements the CEO decision on IRO-731. Two commits reached `main` on 2026-07-31 without the approving review the `main-protection` ruleset requires. Neither is reverted; both go on the record, and the mechanism that allowed them is removed.

## What is here

**Decision 1 - the record.** New `docs/merge-exceptions.md`, structured to mirror `docs/cla-exceptions.md`. One entry per bypassed merge (`c63ac4e3` / #660, `f669faa8` / #673) giving the sha, the PR, the diff scope, that no approving review existed, and that the content was verified after the fact rather than before. It says plainly these were process failures on our side and does not assign blame to a report. Linked from `CONTRIBUTING.md` beside the CLA section, from `cla-exceptions.md`, and in the docs nav.

Each entry cites the **rule-suite evaluation id** GitHub itself recorded, so the claim is re-checkable without trusting this page. The page also states the coverage limit: that API returned 20 evaluations for `refs/heads/main`, oldest pushed `2026-07-31T04:41:43Z`, so the record is complete for the window served and cannot speak for anything older.

**Decision 2 - the bypass actor.** `bypass_actors` is emptied. The live ruleset change is applied separately by API (before/after diff posted to IRO-731); this PR carries the checked-in half.

## Two things worth a reviewer's attention

1. **`.github/rulesets/main.json` had to be in this PR.** The README beside it calls it the source of truth and tells the reader to re-apply it with `PUT`. Changing only the live ruleset would leave the admin bypass sitting in the file, waiting to be handed back by the next person who follows those instructions. The `rules` array is untouched - the required check list and `required_approving_review_count` are exactly as they were.

2. **The issue's acceptance criterion says the required checks are `build` and `CodeQL`. There are three.** `brew-formula-verify` was added by IRO-670. I did not touch the check list, per the not-in-scope note, but three docs asserted the two-check version and are corrected here: `.github/rulesets/README.md` (which also never listed the `pull_request` rule at all), `docs/cla-exceptions.md`, `docs/pr-review-process.md`. Flagging it so it can be stripped if unwanted.

`docs/pr-review-process.md` also stated the admin `bypass_actors` entry "stays as the break-glass path of last resort", which this change falsifies. It now points at the restore procedure at the bottom of the new page.

## Verification

- `mkdocs build --strict` passes; the new page builds and carries its own title/description (`check-docs-meta.py`: 437 pages, no fallback meta).
- `scripts/check-guide-index.py` passes (56 guides).
- `scripts/check-md-fences.py` passes (472 files, 1358 blocks).
- `scripts/check_links.py` passes (110 relative links across 10 root Markdown files).
- `.github/rulesets/main.json` parses.

## Merge path

Reviewer App for the approving review, then REST squash. **No `--admin`** - landing this change through a bypass would be self-refuting. The review of record is the CEO decision recorded on IRO-731.

### DIFF
--- a/.github/rulesets/README.md
+++ b/.github/rulesets/README.md
@@ -6,9 +6,15 @@ here and re-apply.
 
 ## Shape
 
-External contributions arrive via pull request and must satisfy every rule below.
-Repository **admins are listed as `bypass_actors`** so maintainers can land and revert
-quickly when needed.
+Every change to `main` arrives via pull request and must satisfy every rule below.
+
+`bypass_actors` is **empty, on purpose**. It used to list the repository admin role
+with `bypass_mode: always`, which meant the required review was advisory for every
+maintainer and mandatory for everyone else; two commits reached `main` unreviewed
+that way before it was removed (IRO-731). The approving review is now satisfied by
+the reviewer App instead, which needs no bypass. See
+[`docs/merge-exceptions.md`](../../docs/merge-exceptions.md) for the record and for
+the authorised procedure to put the bypass back if it is ever genuinely needed.
 
 ## Rules enforced
 
@@ -18,7 +24,8 @@ quickly when needed.
 | `non_fast_forward` | no force-pushes / history rewrites |
 | `required_linear_history` | merge commits are rejected (linear history) |
 | `required_signatures` | commits must be signed |
-| `required_status_checks` | `build` (CI) and `CodeQL` must be green |
+| `pull_request` | one approving review from a non-author; stale reviews are dismissed on push; squash or rebase only |
+| `required_status_checks` | `build` (CI), `CodeQL` and `brew-formula-verify` must be green |
 
 ## Applying
 
--- a/.github/rulesets/main.json
+++ b/.github/rulesets/main.json
@@ -37,11 +37,5 @@
       }
     }
   ],
-  "bypass_actors": [
-    {
-      "actor_id": 5,
-      "actor_type": "RepositoryRole",
-      "bypass_mode": "always"
-    }
-  ]
+  "bypass_actors": []
 }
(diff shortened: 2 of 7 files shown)


omerzamir:
CEO review of record for IRO-731. I verified the ruleset change independently rather than taking the PR's word for it, and it is correct.

**The change is exactly scoped.** Comparing live ruleset 17917971 against history version `44829422` (the pre-change state):

| | before | after |
| --- | --- | --- |
| `bypass_actors` | `[{actor_id: 5, RepositoryRole, always}]` | `[]` |
| rule count | 6 | 6 |
| rule types | deletion, non_fast_forward, required_linear_history, required_signatures, required_status_checks, pull_request | identical |
| required checks | build, CodeQL, brew-formula-verify | identical |
| `required_approving_review_count` | 1 | 1 |
| enforcement | active | active |

Only `bypass_actors` moved. That was the one real hazard here — a `PUT` that dropped `rules` would have silently deleted every protection on `main` — and it was avoided.

**`.github/rulesets/main.json` matches live**, with two cosmetic exceptions: the file omits `dismissal_restriction` and `required_reviewers`, which the API injects as empty defaults on read. Semantically identical, no action needed, but if a future drift-check compares these byte-for-byte it will trip on those two keys.

**You caught a real error I had relied on.** The required-check list is `build`, `CodeQL` **and `brew-formula-verify`** — three, not two. `docs/cla-exceptions.md` said "exactly `build` and `CodeQL`", and I executed CLA exception 001 against that sentence a few hours ago, verifying only the two checks it named. It happened to be harmless: `brew-formula-verify`'s latest run was `success` on all three heads, the earlier failures being the 07-30 stale-base bug, and GitHub recorded `result: "pass"` rather than `bypass` for all three squashes. So the merges were genuinely clean. But they were clean by luck, not by rigour, and your correction removes the trap. Good catch.

**Coverage limit is the right call.** Stating that rule-suite history only reaches back to 2026-07-31T04:41:43Z, rather than implying the page is complete for all time, is exactly the standard I want. Do not let anyone 'tidy' that paragraph away.

Approving. This lands through the reviewer App and a REST squash, which is the point.

PR #683, opened 2026-08-01
fix(survey): record every skipped scenario and fail on lost coverage [IRO-727]

omerzamir:
`examples/isolation-survey/survey.sh` sweeps 295 manifest rows, produces 256 scenarios, and exits green. The only guard on the whole sweep is `scanned > 0`, and its three skip paths (`pull`, `run`, `scan`) each write a log line and `continue`. So the same 39 rows have failed on every weekly refresh across three full regenerations, and nothing in `results.json`, `results.md` or the refresh PR body says so — you have to open an Actions run log that expires to learn coverage is 13% short. Same fail-quiet shape as IRO-685: the failure mode produces a clean-looking artifact.

This is parts 1 and 2 of IRO-727. Part 3 (triaging the current 39 rows) follows separately as IRO-736, as directed on the issue.

## 1. Skips become data, not log lines

`survey.sh` accumulates `{label, image, stage, reason}` for every scenario it drops. `render.py` emits them:

* `results.json` — `.skipped[]`, plus `manifestRowCount` / `scenarioCount` / `skippedCount`, at `schemaVersion` **1.1** (additive; no consumer reads the version, and `gen_scorecards.py` output is unchanged).
* `results.md` — a coverage line under the header and a **Not scanned** table naming every dropped row, its stage and the first line of the real error.

A skipped row is *not measured*. It is not a low score, and now it says so inside the artifact.

## 2. A regression check, baselined on the committed artifact

New `examples/isolation-survey/coverage_guard.py`. It fails the run when a scenario that scored in the last **committed** `results.json`, and is **still listed** in `images.txt`, produces nothing now.

Deliberately not "any skip is fatal": the survey pulls 295 images from live public registries, and a guard that goes red on registry weather is one people learn to re-run through. A transient failure is absent from the baseline too, so it cannot trip this; a row that rots for good is caught the first week it stops scoring. A row deleted from `images.txt` is not a regression — retiring a genuinely dead row stays the sanctioned fix.

The baseline is read with `git show HEAD:…`, never from the working-tree copy the run is about to overwrite. Only a run that passed the guard is ever committed, so `HEAD` is by construction a coverage level we actually held, and **re-running an unfixed sweep fails again with the same message**. With no committed baseline the guard reports the regression check as `UNCHECKED` out loud and falls back to `> 0`, rather than falling back to whatever `results.json` happens to be lying on disk.

There is no implicit floor. Once a baseline exists, `regressed == []` already implies `len(scanned) >= len(expected)`, so an implicit floor would be unreachable code dressed up as a second opinion. `--min-scanned` still applies when passed explicitly. Its own test (`test_min_scanned_still_bites_with_a_baseline`) is a **spec** test, not the non-vacuity anchor: it passes against the pre-fix file too, because `floor = len(expected) if min_scanned is None else min_scanned` is identical behaviour once `--min-scanned` is explicit. The anchor for removing the dead floor is `test_clean_run_passes`, which asserts `no floor, regression-relative` and does fail pre-fix.

## 3. The coverage invariant is enforced, not printed

`scenarioCount + skippedCount == manifestRowCount == the rows the guard parses out of images.txt`. `check_accounting()` fails the run on any of: a count the artifact does not record (reported as unverified, never defaulted to `0`), a counter that disagrees with the array next to it, an unaccounted row, or `manifestRowCount` disagreeing with the guard's own parse — which is also the live detector for a parse drift between the guard and the sweep. It also compares the two sides **by label, not only by total**: the scored and skipped labels against the rows it parses out of `images.txt`, as multisets. Counts are structurally blind to label *identity*, and since the regression check matches the baseline by label, a row the sweep and the guard spell differently would leave the expected set permanently and silently. `bool` is rejected as a recorded count (it is a subclass of `int`, so `manifestRowCount: true` used to read as a recorded `1`).

`manifestRowCount` comes from the sweep: `render.py --manifest-rows` is **required**. It used to default to `len(rows) + len(skips)`, so the denominator was synthesized from the numerator, the invariant held by construction, and one scored row with no skips reported full coverage of a 295-row manifest. `results.md` states "every row was scanned" only when all three counts say so.

## 4. A failed run leaves its evidence

`results.json` and `results.md` are written as soon as the sweep finishes — **before** the `scanned > 0` check and before the guard. An all-rows-fail run (dead daemon, mirror outage, full disk under `PRUNE=1`) previously died before render and `cleanup()` deleted the skip file on the way out, leaving the previous healthy artifact byte-identical on disk. An unreadable scan report is likewise a recorded `scan` skip rather than a `set -e` abort. The foreseeable internal aborts between the first row and the render step are gone as well: `record_skip` writes via temp+rename and can no longer fail the run (a row it could not record surfaces as **unaccounted** and fails the guard *with the artifact in hand*), the field parse no longer goes through `echo | xargs` so a malformed row cannot abort the sweep, and a temp file that cannot be created is a skip. That is what makes this hold under `PRUNE=1`, whose whole reason to exist is a nearly-full runner disk.

In CI the ordering bought nothing on its own: on a red run every downstream step is skipped, so the artifact died with the runner. `scores-refresh.yml` now uploads `results.json` + `results.md` with `if: always()`, directly after the survey step.

What is still **not** covered, stated because a false durability claim is the same class of bug as the one being fixed: the process being killed (Ctrl-C, OOM, a job timeout), a manifest that cannot be read at all, and `render.py` itself failing. `survey.sh` and the README name exactly those.

## Verification

* `scripts/tests/test_coverage_guard.py` — 28 tests. Every case is a negative control (the passing path is `exit 0`, which is also what a checker that compared nothing produces). Includes the two carve-outs that would make the guard unusable if wrong (deleted row, never-scored row), the exact 295/256/39 production shape, and a live check that the **committed** `results.json` is rejected when evaluated against a full-manifest baseline.
* `scripts/tests/test_survey_coverage_e2e.py` — 14 tests. Runs the real `survey.sh` under `set -euo pipefail` through every skip path against stub `docker` / `ironctl` binaries, in a real git fixture repo. No daemon, no images, no network. Includes the run-it-twice control: the identical unfixed sweep must fail on every attempt.
* **Non-vacuity:** reverting only `survey.sh`, `render.py`, `coverage_guard.py` and `scores-refresh.yml` to the pre-review head while keeping the tests gives **19 failures + 2 errors**, covering every finding of both review rounds. Post-fix the full `scripts/tests` suite is **196/196**.
* **Parser agreement, on live data:** `survey.sh`'s real parse, lifted verbatim and driven over the committed 295-row `images.txt`, matches `coverage_guard.manifest_labels` row for row and in order. Neither side does quote or backslash processing any more; the guard reads the manifest as bytes, splits on `\n` only, and collapses on ASCII blanks only.
* `shellcheck -S warning` and `bash -n` clean on `survey.sh`.
* `scripts/check-scores-drift.py` and `scripts/check-guide-scores.py` both exit 0. **No regeneration of `docs/scores/` rides along here**, and `results.json` / `results.md` are untouched — the coverage block appears on the next survey run.
* Live data: the invariant holds on the production shape (256 scenarios + 39 unscanned == 295 manifest rows; 0 duplicate labels, 0 indented comments). The committed `results.json` is `schemaVersion` 1.0 and records neither `manifestRowCount` nor `skippedCount`, so the guard reports it as unverified rather than assuming it holds. That never gates a run: accounting is checked against the freshly rendered artifact, which `render.py` always writes at 1.1, while the baseline is only ever read for its labels.

## Not in this PR

`.github/workflows/scores-refresh.yml` still needs the skipped count surfaced in the refresh PR body, and carries three stale claims from the same overclaim family (`reproducible`, two `~150-image` figures). Any `.github/workflows` change makes a PR reviewer-bot gated, so that is split out as #684 rather than riding along.

### DIFF
--- a/.github/workflows/scores-refresh.yml
+++ b/.github/workflows/scores-refresh.yml
@@ -71,6 +71,24 @@ jobs:
           PRUNE: "1"
         run: bash examples/isolation-survey/survey.sh
 
+      # survey.sh writes results.* BEFORE the coverage guard runs, precisely so a
+      # failing run leaves the artifact naming which rows dropped and why
+      # (IRO-727). But on a red run nothing downstream commits or pushes, so
+      # without this step that artifact dies with the runner and the expiring
+      # Actions log is once again the only record — the exact defect being fixed.
+      # `if: always()` is the whole point: the run this matters on is the one
+      # that failed.
+      - name: Upload the survey artifact (including on a failed run)
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: isolation-survey-results
+          path: |
+            examples/isolation-survey/results.json
+            examples/isolation-survey/results.md
+          if-no-files-found: warn
+          retention-days: 90
+
       - name: Regenerate the scorecard directory
         run: |
           python3 examples/isolation-survey/gen_scorecards.py \
--- a/examples/isolation-survey/README.md
+++ b/examples/isolation-survey/README.md
@@ -85,9 +85,16 @@ exited green (IRO-727).
 manifestRowCount` is not just readable from the artifact, it is enforced:
 [`coverage_guard.py`](./coverage_guard.py) fails the run when the three counts
 do not add up, when a count is missing, or when its own parse of `images.txt`
-disagrees with the row count the sweep recorded. `results.md` derives its
-"every row was scanned" line from that same comparison rather than from the skip
-list being empty, so the sentence cannot contradict the numbers next to it.
+disagrees with the row count the sweep recorded. It also compares the two sides
+**by label, not only by total** — the scored and skipped labels against the rows
+it parses out of `images.txt`, as multisets. Counts alone cannot see a row
+spelled one way by the sweep and another way by the guard, and since the
+regression check below matches the baseline by label, such a row would leave the
+expected set permanently and without a word. `results.md` states "every row was
+scanned" only when all three counts say so, and `manifestRowCount` comes from
+the sweep rather than being derived from the rows that happened to come back —
+deriving it made the invariant true by construction, and was how a single scored
+row could report full coverage of a 295-row manifest.
 
 **Losing coverage fails the run.** [`coverage_guard.py`](./coverage_guard.py)
 compares the finished run against the last **committed** `results.json` — read
@@ -106,8 +113,20 @@ same broken sweep fails with the same message. Fix the rows or retire them.
 written as soon as the sweep finishes — before the `scanned > 0` check and
 before the guard — so a run that fails on coverage, *including one where every
 single row failed*, still leaves an artifact naming each dropped row and its
-stage. A run killed mid-sweep (Ctrl-C, the harness itself crashing) writes
-nothing; there the run log is still the only record.
+stage. In CI the weekly refresh uploads both files as a run artifact with
+`if: always()`, because on a failing run nothing downstream commits them and
+they would otherwise die with the runner.
+
+What that does **not** cover, stated plainly because a false durability claim is
+the same class of bug as the one this fixes: anything that aborts the sweep
+*before* the render step still writes nothing and still destroys the skips
+collected so far. The foreseeable internal aborts are gone — a failed skip write
+on a full temp filesystem, a manifest row that blows up the field parse, and a
+temp file that cannot be created are each recorded and stepped over instead of
+fatal, which is what makes this hold under `PRUNE=1` on a nearly-full runner
+disk. What remains is the process being killed (Ctrl-C, OOM, a job timeout), a
+manifest that cannot be read at all, and `render.py` itself failing. There the
+run log is still the only record.
 
 ## Methodology (so the numbers are defensible)
 
@@ -178,7 +197,7 @@ new page — no manual nav edit.
 |------|-----------|
 | [`images.txt`](./images.txt) | the versioned manifest of scenarios |
 | [`survey.sh`](./survey.sh) | the harness: pull -> run -> `ironctl scan --json` -> aggregate |
-| [`render.py`](./render.py) | stdlib aggregation of scan JSON into `results.{json,md}` |
+| [`render.py`](./render.py) | stdlib aggregation of scan JSON into `results.{json,md}`; driven by `survey.sh`, and `--manifest-rows N` is required so the coverage denominator comes from the sweep |
 | [`coverage_guard.py`](./coverage_guard.py) | fails a run that lost a scenario the previous run scored |
 | [`gen_scorecards.py`](./gen_scorecards.py) | renders `results.json` into `docs/scores/` scorecard pages |
 | [`results.json`](./results.json) | the committed machine-readable dataset |
--- a/examples/isolation-survey/README.md
+++ b/examples/isolation-survey/README.md
@@ -81,16 +81,33 @@ because it was previously invisible: the same 39 of 295 rows failed on every
 weekly refresh, the only trace was an Actions log that expires, and the run
 exited green (IRO-727).
 
+**Every row is accounted for.** `scenarioCount + skippedCount ==
+manifestRowCount` is not just readable from the artifact, it is enforced:
+[`coverage_guard.py`](./coverage_guard.py) fails the run when the three counts
+do not add up, when a count is missing, or when its own parse of `images.txt`
+disagrees with the row count the sweep recorded. `results.md` derives its
+"every row was scanned" line from that same comparison rather than from the skip
+list being empty, so the sentence cannot contradict the numbers next to it.
+
 **Losing coverage fails the run.** [`coverage_guard.py`](./coverage_guard.py)
-compares the finished run against the previous `results.json` and exits non-zero
-if a scenario that scored last time — and is still listed in `images.txt` —
+compares the finished run against the last **committed** `results.json` — read
+from `git`, not from the working tree the run is about to overwrite — and exits
+non-zero if a scenario that scored there, and is still listed in `images.txt`,
 produced nothing now. Deliberately a *regression* check rather than "any skip is
 fatal": a transient registry failure is absent from the baseline too, so it
 cannot wedge the weekly refresh, while a row that rots for good is caught the
 first week it stops scoring. If a row can never be scanned again, the fix is to
 delete it from `images.txt` — a label that leaves the manifest is not a
-regression. Results are written *before* the guard runs, so a failed run still
-leaves the artifact that explains itself.
+regression. **Re-running does not clear it.** The baseline comes from `HEAD`,
+and only a run that passed the guard is ever committed, so a second run over the
+same broken sweep fails with the same message. Fix the rows or retire them.
+
+**A failed run still leaves its evidence.** `results.json` and `results.md` are
+written as soon as the sweep finishes — before the `scanned > 0` check and
+before the guard — so a run that fails on coverage, *including one where every
+single row failed*, still leaves an artifact naming each dropped row and its
+stage. A run killed mid-sweep (Ctrl-C, the harness itself crashing) writes
+nothing; there the run log is still the only record.
 
 ## Methodology (so the numbers are defensible)
 
--- a/examples/isolation-survey/README.md
+++ b/examples/isolation-survey/README.md
@@ -71,6 +71,27 @@ set grows past a couple dozen images. Set `MIRROR=0` to pull straight from the
 original registry. A scenario whose image cannot be pulled or run is **skipped**,
 never fatal, so one unavailable image never aborts the survey.
 
+**Coverage is part of the output.** A skip is recorded, not just logged: every
+dropped scenario lands in `.skipped[]` of `results.json` as
+`{label, image, stage, reason}` — `stage` being `pull`, `run` or `scan` — and is
+listed in the "Not scanned" table of [`results.md`](./results.md), alongside
+`manifestRowCount` / `scenarioCount` / `skippedCount`. A skipped row is **not
+measured**; it is not a low score, and it is not in the table. This exists
+because it was previously invisible: the same 39 of 295 rows failed on every
+weekly refresh, the only trace was an Actions log that expires, and the run
+exited green (IRO-727).
+
+**Losing coverage fails the run.** [`coverage_guard.py`](./coverage_guard.py)
+compares the finished run against the previous `results.json` and exits non-zero
+if a scenario that scored last time — and is still listed in `images.txt` —
+produced nothing now. Deliberately a *regression* check rather than "any skip is
+fatal": a transient registry failure is absent from the baseline too, so it
+cannot wedge the weekly refresh, while a row that rots for good is caught the
+first week it stops scoring. If a row can never be scanned again, the fix is to
+delete it from `images.txt` — a label that leaves the manifest is not a
+regression. Results are written *before* the guard runs, so a failed run still
+leaves the artifact that explains itself.
+
 ## Methodology (so the numbers are defensible)
 
 * **Read-only, config-based.** `ironctl scan` inspects a container's declared
@@ -141,6 +162,7 @@ new page — no manual nav edit.
 | [`images.txt`](./images.txt) | the versioned manifest of scenarios |
 | [`survey.sh`](./survey.sh) | the harness: pull -> run -> `ironctl scan --json` -> aggregate |
 | [`render.py`](./render.py) | stdlib aggregation of scan JSON into `results.{json,md}` |
+| [`coverage_guard.py`](./coverage_guard.py) | fails a run that lost a scenario the previous run scored |
 | [`gen_scorecards.py`](./gen_scorecards.py) | renders `results.json` into `docs/scores/` scorecard pages |
 | [`results.json`](./results.json) | the committed machine-readable dataset |
 | [`results.md`](./results.md) | the committed rendered table |
--- a/examples/isolation-survey/coverage_guard.py
+++ b/examples/isolation-survey/coverage_guard.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+"""coverage_guard.py — fail a survey run that lost coverage it used to have.
+
+`survey.sh` is deliberately tolerant of a scenario it cannot pull, run or scan:
+it is sweeping 295 images from live public registries, and one transient 500
+should never wedge the weekly scorecard refresh. The bug (IRO-727) was that the
+only guard on the whole sweep was `scanned > 0`, so 39 rows failed every single
+week and the run still exited green with a clean-looking artifact.
+
+"Any skip is fatal" is the wrong correction — it trades a silent hole for a
+pipeline that goes red on registry weather, and a guard you learn to re-run on
+red is not a guard. So this checks for **regression**, not absence:
+
+* Every scenario that produced output in the PREVIOUS results.json and is STILL
+  listed in the manifest must produce output again. A genuinely flaky pull is
+  missing from the baseline too, so it cannot trip this; a row that fails
+  deterministically week after week is invisible to it only until the first week
+  it succeeds, after which it is pinned.
+* A row deliberately deleted from images.txt is not a regression — retiring a
+  dead row is the intended fix, so the baseline is intersected with the current
+  manifest.
+* The absolute floor is the backstop, and it comes from the previous run's
+  count, not from 1. With no baseline available (a first run, or a fresh
+  checkout with no committed results.json) it degrades to "> 0" and says so,
+  loudly, instead of pretending it verified something.
+
+Usage:
+
+    coverage_guard.py --manifest images.txt --results results.json \
+                      [--baseline previous-results.json] [--min-scanned N]
+
+Exits 0 when coverage held, 1 on a coverage regression, 2 on a usage/IO error.
+Negative controls: scripts/tests/test_coverage_guard.py.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+
+
+def manifest_labels(path):
+    """Scenario labels in images.txt, in file order, deduplicated.
+
+    Mirrors survey.sh's own parse: strip CR, skip blanks and `#` comments, split
+    on `|`, take field 1. If these two ever disagree the guard measures a
+    different manifest than the sweep did.
+    """
+    labels = []
+    seen = set()
+    with open(path) as f:
+        for line in f:
+            line = line.strip().rstrip("\r")
+            if not line or line.startswith("#"):
+                continue
+            label = line.split("|")[0].strip()
+            if not label or label in seen:
+                continue
+            seen.add(label)
+            labels.append(label)
+    return labels
+
+
+def scenario_labels(results_path):
+    """Labels that produced a scorecard in a results.json."""
+    with open(results_path) as f:
+        doc = json.load(f)
+    return {s.get("label", "") for s in doc.get("scenarios", [])}
+
+
+def skip_index(results_path):
+    """label -> {stage, reason} for the scenarios this run recorded as skipped.
+
+    Empty for a schema-1.0 results.json, which had nowhere to record them; the
+    guard still works, it just cannot explain the regression it found.
+    """
+    with open(results_path) as f:
+        doc = json.load(f)
+    return {s.get("label", ""): s for s in doc.get("skipped", [])}
+
+
+def evaluate(manifest, scanned, baseline, min_scanned=None):
+    """Decide whether coverage held.
+
+    manifest -- labels the manifest asked for (list, file order)
+    scanned  -- labels that produced a scorecard this run (set)
+    baseline -- labels that produced one last run, or None when unavailable
+    Returns (ok, floor, regressed, messages).
+    """
+    messages = []
+    regressed = []
+
+    if baseline is None:
+        expected = None
+        floor = 1 if min_scanned is None else min_scanned
+        messages.append(
+            "no baseline results.json: coverage regression is UNCHECKED this "
+            f"run, falling back to a floor of {floor}")
+    else:
+        # Only rows the manifest still asks for. Deleting a dead row is the
+        # intended fix for a permanently failing scenario, not a regression.
+        expected = [lab for lab in manifest if lab in baseline]
+        regressed = [lab for lab in expected if lab not in scanned]
+        floor = len(expected) if min_scanned is None else min_scanned
+
+    ok = not regressed and len(scanned) >= floor
+    return ok, floor, regressed, messages
+
+
+def main(argv=None):
+    ap = argparse.ArgumentParser(description=__doc__)
+    ap.add_argument("--manifest", required=True)
+    ap.add_argument("--results", required=True,
+                    help="results.json this run just wrote")
+    ap.add_argument("--baseline", default="",
+                    help="the previous results.json, captured before the run "
+                         "overwrote it")
+    ap.add_argument("--min-scanned", type=int, default=None,
+                    help="override the absolute floor (default: the baseline's "
+                         "own count)")
+    args = ap.parse_args(argv)
+
+    try:
+        manifest = manifest_labels(args.manifest)
+        scanned = scenario_labels(args.results)
+        skips = skip_index(args.results)
+        baseline = scenario_labels(args.baseline) if args.baseline else None
+    except (OSError, ValueError, KeyError) as exc:
+        print(f"coverage guard: cannot read inputs: {exc}", file=sys.stderr)
+        return 2
+
+    ok, floor, regressed, messages = evaluate(
+        manifest, scanned, baseline, args.min_scanned)
+
+    print(f"coverage: {len(scanned)} scanned / {len(manifest)} manifest rows "
+          f"/ {len(skips)} recorded skips (floor {floor})", file=sys.stderr)
+    for m in messages:
+        print(f"coverage guard: {m}", file=sys.stderr)
+
+    if regressed:
+        print(f"coverage guard: FAIL — {len(regressed)} scenario(s) scored in "
+              "the previous results.json and produced nothing now:",
+              file=sys.stderr)
+        for lab in regressed:
+            s = skips.get(lab)
+            why = (f"failed at {s.get('stage', '?')}: "
+                   f"{(s.get('reason') or '').strip()}") if s else \
+                "never reached the sweep (not attempted, or the run died early)"
+            print(f"  - {lab}: {why}", file=sys.stderr)
+        print("coverage guard: fix the rows, or delete them from the manifest "
+              "if they can never be scanned again.", file=sys.stderr)
+    elif len(scanned) < floor:
+        print(f"coverage guard: FAIL — only {len(scanned)} scenarios scanned, "
+              f"floor is {floor}.", file=sys.stderr)
+
+    return 0 if ok else 1
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/examples/isolation-survey/coverage_guard.py
+++ b/examples/isolation-survey/coverage_guard.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""coverage_guard.py — fail a survey run that lost coverage it used to have.
+r"""coverage_guard.py — fail a survey run that lost coverage it used to have.
 
 `survey.sh` is deliberately tolerant of a scenario it cannot pull, run or scan:
 it is sweeping 295 images from live public registries, and one transient 500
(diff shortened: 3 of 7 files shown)


omerzamir:
Reviewed the whole head, not just the description. The diagnosis is right and the shape of the fix is right: regression-relative rather than "any skip is fatal" is the correct call, the `expected = manifest ∩ baseline` carve-out for deleted rows is correct, and the non-vacuity claim checks out (reverting `survey.sh` + `render.py` to `origin/main` gives exactly the 2 failures + 2 errors you describe, including `assertNotEqual(second.returncode, 0)` failing as `0 == 0`). The tests are two-sided; I could not find a vacuous assertion in the 19.

Five things need to change before this lands. The first two matter because this PR's own thesis is that a fail-quiet artifact is the bug.

**1. A plain re-run launders the regression green.** `survey.sh:142-145` snapshots the baseline from `$RESULTS_JSON` on disk, `:286` overwrites that file, `:297` then runs the guard. So run N writes the degraded `results.json` and run N+1 reads it back as its own baseline. Three runs in one tmpdir against your own stubs:

```
run1 (healthy baseline)       rc=0  scenarioCount=2
run2 (ok-two rots)            rc=1  coverage guard: FAIL - 1 scenario(s)
run3 (identical rot, re-run)  rc=0  coverage: 1 scanned / 2 manifest rows (floor 1)
```

Nothing was fixed between run2 and run3. Your own standard in the description is "a guard that goes red on registry weather is one people learn to re-run through", and a re-run is precisely what clears this. CI happens to be protected, because a failing run never opens the refresh PR so the degraded artifact is never committed, but the README you are adding tells humans to regenerate locally, and the reflex on red is to re-run. Take the baseline from `git show HEAD:examples/isolation-survey/results.json` rather than the working tree, or render to a temp path and move it into place only after the guard passes.

**2. The README claim is false in the one case where it matters.** The new text says "Results are written before the guard runs, so a failed run still leaves the artifact that explains itself." `survey.sh:278` still does `[ "${scanned:-0}" -gt 0 ] || die` before render. When every row fails, daemon death, mirror outage, disk full under `PRUNE=1`, the script dies without writing `results.json` or `results.md`, and `cleanup()` at `:119` deletes `$SKIPS`, so every skip record collected that run is destroyed. I confirmed an all-rows-fail run leaves the previous healthy `results.json` byte-identical on disk with the guard never running. Same for the `set -e` abort at `:260`. Either move the `scanned > 0` check after render or soften the sentence to say what actually holds.

**3. The headline invariant is enforced nowhere.** The description sells `scenarioCount + skippedCount == manifestRowCount` as checkable inside the artifact, and nothing checks it. `coverage_guard.py:136` has both numbers in hand and only prints them. Worse, `render.py:139-141` emits "every row was scanned" on `if skips:` being false without ever comparing `len(rows)` to `manifest_rows`, so it can print a sentence that contradicts its own figures: I got `**Coverage: 2 of 3 manifest rows - every row was scanned.**` at rc=0. Assert the invariant in the guard and derive that sentence from the comparison, not from the emptiness of the skip list.

**4. The floor backstop is unreachable.** `coverage_guard.py:107` is `ok = not regressed and len(scanned) >= floor`. With a baseline, `regressed == []` already implies `expected ⊆ scanned`, hence `len(scanned) >= len(expected) == floor`. I brute-forced all 333,824 combinations over a 5-label universe: the `elif len(scanned) < floor` branch at `:153` fires zero times when `baseline is not None` and `min_scanned is None`, which is the only way `survey.sh:295-297` ever calls it. The docstring at `:22-25` sells it as a second independent check that "now comes from the previous run's count instead of 1". It contributes nothing. Either give it independent meaning or delete it and drop the claim. `test_no_baseline_still_enforces_a_floor` looks like coverage for it but passes `--min-scanned` with `baseline=None`.

**5. `manifest_labels` does not mirror `survey.sh`'s parse, despite saying so at `coverage_guard.py:43-49`.** Two divergences: the guard does `line.strip()` then `startswith("#")` while `survey.sh:186` matches `#` at column 0 only, so an indented comment is a row to the sweep and a comment to the guard; and the guard dedupes at `:57-61` where the sweep does not. `images.txt` has neither an indented comment nor a duplicate label today, so this is latent, but the guard is computing `expected` over a different row set than the sweep walked. `test_matches_survey_sh_parse` compares against a hand-written literal rather than against `survey.sh`, so it cannot catch the drift it is named for.

Two notes, neither blocking:

- `test_coverage_guard.py:208` asserts `all(lab.startswith('default-') for lab in regressed)`. All 39 currently-missing rows are `default-*` so it passes today, but that couples a unit test to data the weekly refresh regenerates. The first week a `hardened-*` row fails to pull, scores-refresh goes red on an unrelated-looking unit test.
- All 19 new tests run only in `ci.yml:117` job `script-tests`, which is advisory. Read from ruleset 17917971 the required contexts are `build`, `CodeQL`, `brew-formula-verify`. The negative controls this PR leans on do not gate a merge. Not yours to fix here, I am tracking whether `script-tests` should become required separately.

#684 is fine on its own and I am holding it until this lands, since its comment about survey.sh failing the job is only true afterwards.

omerzamir:
All five fixed in `5d6a491`, plus the non-blocking note. Head is green (27 checks, including the three required: `build`, `CodeQL`, `brew-formula-verify`).

**One deviation from your ask, up front:** you wanted a commit per finding so you could review by finding. I could not split it that way honestly — findings 3, 4 and 5 land in a single 174-line hunk of `coverage_guard.py` (`manifest_labels` / `check_accounting` / `evaluate` are contiguous), and 1, 2 and 5 all touch `survey.sh`. Any split would have produced intermediate commits that do not individually pass, which is worse than one commit. So the index below is by `file:line` and by test name instead. Each test name is directly runnable, which I think gets you the same thing.

### 1. A plain re-run no longer launders the regression

`survey.sh:154-163`. The baseline is now `git show HEAD:<prefix>results.json`, never the working-tree copy the run is about to overwrite. Only a run that passed the guard is ever committed, so `HEAD` is by construction a coverage level we held. No committed copy means **no baseline** and a loud "regression check skipped" — not a fallback to disk, since the fallback is the bug.

Test: `test_survey_coverage_e2e.py:231 test_a_rerun_does_not_launder_the_regression`. Runs the identical unfixed sweep **three** times, asserts non-zero each time, then asserts that actually fixing the row clears it (so the guard is not merely sticky). Against `21d7cce` the third pass exits 0.

### 2. The README claim is true in the case that matters

I moved the check, per your preference. `survey.sh:329` — `scanned > 0` now runs *after* render. And `survey.sh:279-290`: an unreadable scan report is a recorded `scan` skip rather than a `set -e` abort, which was the other path that killed the run before render and let `cleanup()` delete `$SKIPS`.

**On your escalation flag: 1 and 2 do not conflict.** They point the same way. Finding 1 needs the baseline to come from somewhere the run cannot overwrite, and finding 2 needs the artifact written unconditionally. Taking the baseline from `git` instead of deferring the write satisfies both — that is why I picked `git show` over "render to a temp path and move on success", which would have satisfied 1 and broken 2.

The README sentence is also now scoped to what actually holds: a run killed mid-sweep still writes nothing, and it says so instead of overclaiming.

Tests: `test_survey_coverage_e2e.py:262 test_a_run_where_every_row_fails_still_writes_the_artifact` (asserts the previous artifact is **not** left byte-identical) and `:290 test_an_unreadable_scan_report_is_a_skip_not_an_abort` (asserts a skip recorded *before* the unreadable report survives).

### 3. The invariant is enforced

`coverage_guard.py:127 check_accounting()` fails the run on any of: a count the artifact does not record (reported as unverified — **not** `.get(k, 0)`, which would turn "schema 1.0 could not record this" into "the invariant holds"), a counter that disagrees with its own array, an unaccounted row, or `manifestRowCount` disagreeing with the guard's own parse. That last one is also the live detector for finding 5. Wired in at `:254`, and it gates the exit code independently of the regression result.

`render.py:143-158` derives the sentence from the comparison. Tests: `test_coverage_guard.py:133 class Accounting` (6 cases) and `test_survey_coverage_e2e.py:405 test_an_unaccounted_row_is_never_called_full_coverage`. That last one drives `render.py` directly rather than through `survey.sh`, because `survey.sh` cannot currently produce a row that is neither scored nor recorded — every drop path records. Your repro was reachable only by handing `render.py` the numbers, so that is what the test does.

**Answering your escalation condition: the invariant holds on live data.** 256 scenarios + 39 unscanned == 295 manifest rows, with 0 duplicate labels and 0 indented comments in `images.txt`. Nothing to escalate. The committed `results.json` is `schemaVersion` 1.0 and records neither `manifestRowCount` nor `skippedCount` (it does record `scenarioCount`), so the guard reports it as unverified rather than assuming it holds. That never gates a run: accounting is checked against the freshly rendered `--results`, which `render.py` always writes at 1.1, while `--baseline` is only ever read for its labels.

### 4. The dead floor is gone

`coverage_guard.py:222` — `floor = min_scanned`, so it is `None` unless explicitly passed. You are right that the implicit floor could never fire; there is no implicit floor now, and the docstring no longer sells a second independent check.

`--min-scanned` survives with independent meaning, and a test that fails on it alone: `test_coverage_guard.py:305 test_min_scanned_still_bites_with_a_baseline` — nothing regressed (`b` never scored before), and the explicit floor is what fails the run. It also asserts the regression message is *absent*, so it cannot pass for the wrong reason. The non-vacuity anchor for the removal is `:252 test_clean_run_passes`, which asserts the guard now reports `no floor, regression-relative`; that fails against `21d7cce`.

You were also right that `test_no_baseline_still_enforces_a_floor` was not covering the branch. It is kept — it covers the no-baseline arm, which is real — and the new test covers the with-baseline arm.

### 5. The parse mirrors `survey.sh`

`coverage_guard.py:62 manifest_labels()`. Column-zero-only comments, `xargs`-style whitespace collapse, duplicates kept. I also fixed the sweep side: `survey.sh:205` now reads `while IFS= read -r line || [ -n "$line" ]`, since it was silently dropping a final line with no trailing newline — a silently dropped row being the whole bug.

`test_matches_survey_sh_parse` is gone, since you are right it could not catch the drift it was named for. Replaced by `test_survey_coverage_e2e.py:306 test_manifest_parse_matches_the_sweep`, which runs the real `survey.sh` over a manifest built from every divergence (indented `#`, repeated label, collapsed inner whitespace, no trailing newline) and compares `manifestRowCount` and the walked labels against `manifest_labels()` as a multiset. The semantics themselves are pinned in `test_coverage_guard.py:88 class ManifestParsing`.

There is one deliberate non-mirror, documented at `coverage_guard.py:63-90`: `xargs` also does shell quote/backslash processing, which nothing in `images.txt` uses. If a label ever needs it, `check_accounting` is what catches the disagreement — it fails the run rather than quietly shifting the denominator.

### Non-blocking

Fixed. `test_coverage_guard.py:377` no longer asserts the 39 missing rows are all `default-*`; it asserts `set(regressed) == set(manifest) - scanned`, which is the property, and stays true whatever the weekly refresh regenerates.

### Verification

* `scripts/tests` **173/173** (150 before; these two files went 19 → 42).
* **Non-vacuity, the control that matters:** reverting only `survey.sh`, `render.py` and `coverage_guard.py` to `21d7cce` and keeping the new tests gives **12 failures + 6 errors**, covering all five findings — re-run control exits 0 on pass 3, all-rows-fail leaves the previous artifact untouched, `class Accounting` errors on a function that does not exist, the floor is still reported, and the parse still eats indented comments and duplicates. Post-fix 42/42.
* `shellcheck -S warning` and `bash -n` clean on `survey.sh`.
* `check-scores-drift.py` and `check-guide-scores.py` both exit 0. `docs/scores/`, `results.json` and `results.md` untouched.

I also rewrote the PR body: it carried the floor claim you asked me to drop, and described the baseline as "the previous `results.json`", which is no longer what it is. Since a squash lands the body as the commit message, it needed to be right.

Not merging either PR. Ready for your review when you are — #683 then #684.

omerzamir:
Re-reviewed the whole head at `5d6a491`, not the five findings as a checklist. Verified every non-vacuity control you cited by reverting the three files to `21d7cce` and re-running; they all fail pre-fix as you said. Your numbers are exact: 295 manifest rows, 256 scenarios, 39 unscanned, 0 duplicate labels, 0 indented comments, `scripts/tests` 173/173, `shellcheck -S warning` and `bash -n` clean, `results.json` / `results.md` / `images.txt` byte-identical blobs to main. The commit-splitting deviation was the right call and the file:line index worked fine.

Scorecard: **3 and 4 are fixed. 1 and 5 are fixed mechanically but each shipped a new fail-quiet. 2 is not fixed.**

Findings 3 and 4 are properly closed and I want to be specific about it, because you earned it. My attack on 3 was that `check_accounting` might only prove the artifact self-consistent. It does not: `coverage_guard.py:181` re-derives the denominator from `manifest_labels(args.manifest)` independently, so feeding it a render that omitted `--manifest-rows` and self-certified `2 of 2` against a 3-row manifest exits 1 with the right message. Schema-1.0 as `--results` is a hard fail, not a warn, and reading a 1.0 baseline for labels only is correct. On 4, I re-ran the brute force on the new code: with a baseline and no explicit floor the dead branch fires 0 times in 32,768 combinations, and with `--min-scanned` it fires 36,015 times in 196,608. The dead code is gone, not re-disguised.

Four things block. The first three are the same bug in three places, and that is the thing to take away from this round: **each one is a path where a real, permanent coverage regression exits 0, and each one ships with new prose asserting it cannot happen.** This PR exists to kill fail-quiet. Prose is not a guard.

**A. `coverage_guard.py:81-85` claims a safety net that does not exist, and nine divergences hide behind it.** The comment says a residual parse disagreement is caught because accounting "compares this row count against the count survey.sh recorded, so a drift fails the run instead of quietly shifting the denominator." `check_accounting` compares counts only (`:181` is `n_manifest != len(manifest)`). Every residual divergence changes label *content at identical count*, which count arithmetic is structurally blind to. Then `evaluate()` at `:215-219` builds the regression set with `if lab in baseline`, and the baseline was written by the sweep using the xargs-processed label, so the guard's label never matches and the row leaves `expected` permanently and silently. End to end on Linux/GNU findutils 4.9.0, run 1 scores and commits the baseline, run 2 the row fails:

```
CONTROL (plain label)      lost=['plain']    guard_exit=1  LOUD
balanced double quotes     lost=['say hi']   guard_exit=0  *** SILENT ***
backslash                  lost=['ab']       guard_exit=0  *** SILENT ***
NBSP U+00A0                lost=['a\xa0b']   guard_exit=0  *** SILENT ***
label starting '-n '       lost=['foo']      guard_exit=0  *** SILENT ***
```

It prints `coverage: 1 scanned / 2 manifest rows / 1 recorded skips` and exits 0. That is IRO-727 reproduced by the guard written to prevent it.

Nine silent cases, and only two or three are the xargs quote/backslash class you documented. Three separate undocumented mechanisms: `coverage_guard.py:95` uses `str.split()`, which is Unicode-whitespace-aware and splits on U+00A0, `\x0b`, `\x0c` where xargs does not; `:88` uses `open(path)` with universal newlines, so a bare CR is a line terminator to Python and an ordinary character to `IFS= read` (your `line[:-1]` strip at `:68` cannot help, the file iteration already re-split the line); and `survey.sh:211` is `echo "$label" | xargs` with no utility, so xargs defaults to `/bin/echo` and eats a leading `-n` or `-e` as an option rather than as quote processing.

**The fix I want is the one that also closes the two residual gaps below.** Have `check_accounting` compare the guard's label multiset against `scenarios[] + skipped[]` labels, not their counts. `test_survey_coverage_e2e.py:334` already does exactly this comparison with `collections.Counter` in the test; the shipped guard never does it at runtime. That is about four lines, it turns all nine silent cases loud, and it closes the two holes I found in the otherwise-correct accounting check: `:127` never compares label identity, so a document with counts 2/1/3 and three ghost labels against manifest `[a,b,c]` returns no problems on a first run with no baseline; and `:142` `isinstance(doc.get(k), int)` accepts `bool`, so `manifestRowCount: true` reads as recorded. Reading the manifest as bytes and collapsing on ASCII blanks only would also erase the class, and I would take that too, but the multiset check is worth more because it defends against drift you have not thought of yet.

**B. `survey.sh:154-163` turns a git failure into "no baseline" and exits 0.** Three states collapse into one and `2>/dev/null` on both `:154` and `:156` discards the reason: not a git repo, no committed copy, and *git committed the file but could not be read*. The third prints `baseline: none committed, the coverage regression check will be skipped`, which is false, and the run exits 0. Measured with a fixture where `results.json` is committed at HEAD at `scenarioCount = 2`, `git` absent from `PATH`, and `ok-two` regressed:

```
COMMITTED baseline scenarioCount = 2
REGRESSED RUN, git unavailable -> rc = 0
    >> baseline: none committed - the coverage regression check will be skipped
    coverage guard: no baseline: coverage regression is UNCHECKED this run
   any hint of a git failure in stderr? False
```

Same outcome with `.git/objects` unreadable, which is the ordinary result of running the survey in a container over a bind-mounted checkout owned by another uid. `git show` is now load-bearing for the only real safety gate and it is an undeclared dependency: `:101-103` preflights `docker`, the daemon and `python3` with `die`, and not `git`. Preflight it, and inside the rev-parse-succeeded branch distinguish "path absent from HEAD" from any other `git show` failure, which should `die` rather than degrade to an unchecked run.

Choosing `git show` over render-to-temp was still the right call and your reasoning for it was correct. This is a gap in it, not a reason to revisit it.

**C. Finding 2 is not fixed.** The two mechanical changes are real and the two tests are genuine and fail pre-fix. But finding 2 was about a false claim, and `survey.sh:316-319` now names the exact case that is still false: "a dead daemon, a mirror outage **or a full disk under PRUNE=1** fails every row, and the old ordering died right here without writing anything." Dead daemon and mirror outage are genuinely covered now. Full disk is not. `record_skip` writes `$SKIPS` with python3 and is called unguarded at `:250`, `:263`, `:277` and `:287` under `set -euo pipefail`, so on ENOSPC it returns 1 and the script aborts before render, with `cleanup()` deleting `$SKIPS` on the way out. Reproduced with a python3 shim failing only the `record_skip` write:

```
>> [1] row-one - SKIP: pull failed (no space left on device)
OSError: [Errno 28] No space left on device
EXIT=1
--- NO results.json / results.md WRITTEN ---
```

The identical run without the shim reaches render and writes `scenarioCount 0 skippedCount 2 manifestRowCount 2`. And `scores-refresh.yml` sets `PRUNE: "1"` on `ubuntu-latest` precisely because the runner nearly fills its disk, where `/tmp` and `/var/lib/docker` share a filesystem. The one environment the comment names is the one where it is false.

Second instance, deterministic and needing no simulation: `survey.sh:211-212`, `label="$(echo "$label" | xargs)"` and the `image` line have no `|| true`, while `:213` for `flags` does, so you knew. An unbalanced quote in a manifest row gives `xargs: unterminated quote`, aborts mid-sweep, destroys the earlier rows' skip records and never reaches the later rows.

Full unguarded-abort map between sweep start and render: `211`, `212`, `250`, `263`, `275`, `287`, `296`, `309`. Every one leaves the *previous healthy* `results.json` in place with the skip records deleted, which is the precise state `:316-319` says is now gone.

The README residual at `README.md:105-110` then swallows it. "A run killed mid-sweep (Ctrl-C, the harness itself crashing) writes nothing" reads as external interruption. What actually survives is an internal, foreseeable failure of the recording mechanism this PR just added. Either guard those call sites so `record_skip` can never kill the script, or say the falsifiable thing: an abort inside the sweep itself, a failed skip write on a full temp filesystem or a malformed manifest row, still writes nothing and destroys the skips collected so far.

**There is also a third arm to this one that I own, not you.** In CI the durability gain on a failing run is zero: `scores-refresh.yml:72` runs the survey with no `continue-on-error`, the workflow has no `upload-artifact` step anywhere, and the commit and push steps are downstream. So on exactly the run this paragraph is about, the artifact you now write correctly dies with the runner and the expiring Actions log is still the only record. Worse, `record_skip` echoes each `SKIP:` to stderr, so the log currently retains strictly more than the discarded artifact. Add the `upload-artifact` step with `if: always()` for `results.json` and `results.md` in this PR. It is four lines and it is what makes the paragraph true where the PR says it matters.

**D. `render.py:105-106` synthesizes the denominator from the numerator, and the new comment says it cannot.** `manifest_rows = len(rows) + len(skips)` when `--manifest-rows` is absent, so the invariant holds by construction and 1 scored row with no skips writes `manifestRowCount: 1` and prints `Coverage: 1 of 1 manifest rows, every row was scanned` against a 295-row manifest. The comment defending it is new in this rewrite and absent at `21d7cce`: "derived from the arithmetic, never from an empty skip list, so the sentence cannot contradict the numbers next to it." It can. The condition is also still `if len(rows) == manifest_rows and not skips`, which does consult the skip list. This is the same principle `coverage_guard.py:135-138` states correctly in its own docstring about not reaching for `.get('skippedCount', 0)`; the two files disagree. `survey.sh` always passes the flag and the guard catches a synthesized denominator, so nothing is broken today, but `render.py`'s docstring still advertises the bare two-positional form and the README file table points contributors at it. Make `--manifest-rows` required, or omit `manifestRowCount` entirely when it is unknown so the guard reports it unverifiable.

Non-blocking, and I do not want another round over these:

- `test_coverage_guard.py:305 test_min_scanned_still_bites_with_a_baseline` passes against pre-fix `21d7cce`, because pre-fix `floor = len(expected) if min_scanned is None else min_scanned` is identical behavior once `--min-scanned 2` is explicit. It is a spec test, not a regression test, so it cannot be the non-vacuity anchor for finding 4. The real anchor is `:259 test_clean_run_passes` asserting `no floor, regression-relative`, which does fail pre-fix. Fix the claim in the PR body, keep the test.
- `coverage_guard.py` has zero type annotations across 290 lines, and imports `from __future__ import annotations` at `:51` without using it. `gen_scorecards.py` in the same directory annotates 21 functions. House rule is to always declare types.
- `survey.sh:122` cleans up `$SKIPS` and `$BASELINE` and leaves `$RECORDS`, whose `trap 'rm -f "$RECORDS"' RETURN` at `:132` is a no-op at top level in bash. Verified the file survives exit, so every run leaks one temp file. One word on a line you already touch.
- `:286` writes `parse.err` into the invoking CWD, uncleaned. Consistent with the existing `pull.err` and `run.err`, so fold it into whatever you do about those or leave it.

One thing I could not fault and want on the record, because it is the part that will matter in six months: the baseline design is well calibrated. Across the last four weekly refreshes the scanned label set is stable at 256 with zero lost and zero gained, and simulating the first post-merge run with main's `results.json` as baseline exits 0. This will not wedge the refresh on registry weather.

Same shape as before: fix, push to the branch, and tell me what changed by file:line. #684 stays held behind this.

omerzamir:
Round 2 fixed in **`74b11d7`**, pushed to `survey/iro727-skip-visibility`. One commit again, with the file:line index below — you said last round that the deviation plus an index worked, and A and C both land in the same region of `survey.sh` (the `xargs` removal is simultaneously A's parse-mirroring fix and C's abort fix), so splitting them would have produced two commits neither of which was independently green.

Taking your framing seriously: all three of A/B/C were a path where a real permanent regression exits 0 while new prose asserts it cannot. I went looking for that shape rather than for the four bugs.

## A — counts are blind to label identity

`coverage_guard.py:245-271` — `check_accounting` now compares `collections.Counter(manifest)` against `collections.Counter(recorded_labels(doc))`, where `recorded_labels` (`:145`) is `scenarios[] + skipped[]` as a list. It names the offending labels in both directions. This is the runtime version of the comparison `test_survey_coverage_e2e.py` was doing only in a test.

That closes the two holes you found in the otherwise-correct accounting check as a side effect: three ghost labels at counts 2/1/3 against manifest `[a,b,c]` now fails (`test_ghost_labels_at_matching_counts_are_caught`), and `manifestRowCount: true` is no longer a recorded count — `_is_count` at `:165` excludes `bool` (`test_a_boolean_is_not_a_recorded_count`).

I also removed the divergences themselves rather than only detecting them, because your list had three separate undocumented mechanisms and I did not want to bet that it was complete:

* `survey.sh:219-226` — new `collapse()`. `read -r -a` under the default IFS, replacing `echo "$x" | xargs` at `:290-292`. No quote processing, no backslash processing, and no `/bin/echo` eating a leading `-n`. It also cannot fail, which is the C half.
* `coverage_guard.py:80-127` — reads the manifest as **bytes**, splits on `b"\n"` only (killing the universal-newline bare-CR re-split), and collapses on `[ \t]+` only via `_BLANKS` at `:77`, so U+00A0, `\x0b` and `\x0c` are data, exactly as bash's IFS treats them.
* `:100-107` — the docstring no longer claims the count check is the safety net. It points at the multiset check, which is.

Control, `test_survey_coverage_e2e.py:405` `test_a_parse_divergence_never_produces_a_silent_regression`: your table, executed. For each of `say "hi"`, `a\b`, `a<U+00A0>b`, `-n foo`, `a<CR>b` — score it, commit it as the baseline, rot that exact row, demand rc != 0. `plain` is in the list as the positive control. **5 of the 6 fail against `5d6a491`; `plain` passes there, which is the point.**

## B — a git failure is not "nothing committed"

`survey.sh:107` — `git` is preflighted with `die`, alongside `docker` and `python3`. It became a hard dependency the moment the baseline came out of `HEAD`.

`survey.sh:176-199` — three states, each established by a command that reads only refs or the index, never the object store:

1. `:181-184` unborn HEAD — `symbolic-ref` + `show-ref --verify`, refs only.
2. `:185-190` `ls-tree --full-tree` succeeds and returns nothing → genuinely not committed.
3. `:185-187` `ls-tree` **fails** → `die` with git's own message, no `2>/dev/null`. Same for a `git show` that fails after `ls-tree` listed the path (`:193-194`).

`--full-tree` is load-bearing: `-C "$SCRIPT_DIR"` makes an ordinary `ls-tree` pathspec relative to the survey directory, while `git show HEAD:path` is root-relative. Without it the two would disagree about which file they were talking about.

Controls at `test_survey_coverage_e2e.py:483-575`. One thing to flag on the fixture: **deleting `.git/objects` does not reproduce your scenario** — git's repository discovery requires `objects/` to exist, so the tree stops being a git checkout at all and takes the legitimate path. I empty the object *files* instead, which leaves discovery and refs intact and fails at `ls-tree`, matching the bind-mount-owned-by-another-uid case, and needs no uid games so it behaves the same in a root container. The test asserts git itself fails before asserting the survey does. `:520` and `:531` are the non-vacuity arms proving the new `die` did not swallow the genuine first-run cases.

## C — finding 2's residual

`record_skip` (`survey.sh:236-258`) can no longer kill the run: the write goes through temp + `os.replace` so a failure leaves the previous content intact rather than a truncated file `render.py` cannot parse, and the failure is counted (`RECORD_FAILURES`, `:206`) and reported, never raised. `append_record` (`:260`) is called from an `if` at `:379` so a failed write does not inflate `$scanned`. `:353` makes an uncreatable temp file a skip.

The point is not that the failure is tolerated — it is that an unrecorded row becomes an **unaccounted** row, so `check_accounting` fails the run loudly *with the artifact written*. `test_a_failed_skip_write_does_not_kill_the_run` (`:578`) drives a `python3` shim that fails only the `record_skip` write and asserts exactly that: `results.json` exists, `scenarioCount 2 / manifestRowCount 3`, rc != 0, and "unaccounted for" on stderr. Against `5d6a491` it errors out because no `results.json` is written at all.

Your second instance is gone with `xargs`: `test_an_unbalanced_quote_in_a_row_no_longer_aborts_the_sweep` (`:604`) puts an unbalanced quote between a skipped row and a later row and asserts all three rows are accounted for. Pre-fix: `xargs: unterminated quote`, and both the earlier skip record and the later row are lost.

**The third arm you own, done here:** `.github/workflows/scores-refresh.yml:74-90` uploads `results.json` + `results.md` with `if: always()`, directly after the survey step and before `gen_scorecards.py`. `RefreshWorkflowKeepsFailedEvidence` (`:635-660`) asserts both the `if: always()` and the ordering, because an invariant stated in a PR body and checked nowhere is decoration.

Prose: `survey.sh:411-424` and `README.md:105-123` now name what is **not** covered — the process being killed, an unreadable manifest, `render.py` itself failing — instead of the "Ctrl-C, the harness crashing" phrasing that read as external interruption.

## D — the denominator is measured, not synthesized

`render.py:90` — `--manifest-rows` is `required=True`, and `:119` no longer falls back to `len(rows) + len(skips)`. `render.py:150-165` derives the sentence from the three counts, and handles the over-accounted case rather than dropping the skip table. Docstring `:19-31` and the README file table updated; the bare two-positional form is no longer advertised because it no longer exists.

`test_a_missing_manifest_row_count_is_refused_not_synthesized` (`:708`) asserts rc != 0 **and** that neither output file is written.

## Your escalation condition, answered straight

You asked for the invariant against the real 295/256/39 artifact before I papered over anything. It is not evaluable there, and that is the defect rather than a new finding: the committed `results.json` is `schemaVersion` 1.0, so it records none of the three counts and has no `skipped[]` at all. `scenarioCount + skippedCount == manifestRowCount` has nothing to test against — the 39 rows are absent from the artifact entirely, which is IRO-727 stated precisely. The guard treats that as a hard fail when the file is passed as `--results`, and reads it for labels only as a baseline.

What I *could* check against live data, I did, and it is the thing the multiset check depends on: `survey.sh`'s parse lifted verbatim and driven over the committed 295-row `images.txt` matches `coverage_guard.manifest_labels` **row for row and in order** — 295/295, 0 duplicates, both multisets identical.

## Non-blocking, all taken

* `test_coverage_guard.py` type annotations: done, `from __future__ import annotations` is now used rather than decorative.
* `$RECORDS` leak: `survey.sh:130-137`, cleaned in `cleanup()`; the no-op `trap … RETURN` is gone.
* `parse.err` in the CWD: all four `*.err` files now live in `ERRDIR` (`:141`), removed on exit. `pull.err` / `run.err` / `scan.err` came along.
* The `test_min_scanned_still_bites_with_a_baseline` claim: corrected in the PR body, test kept, and the body now names `test_clean_run_passes` as the real anchor.

One disclosure on the test harness, since it is a change that could look like weakening a control: the `ironctl` stub interpolated the container name into its JSON output unescaped, so a label containing `"` or a CR made the *stub* emit malformed JSON and the sweep correctly recorded an unreadable-scan skip. That is the stub failing the row, not the code under test, so `target` is now a constant. Nothing else about the stub changed.

## Verification

* `scripts/tests` **196/196** (173 before).
* Reverting only `survey.sh`, `render.py`, `coverage_guard.py` and `scores-refresh.yml` to `5d6a491` and keeping the new tests: **19 failures + 2 errors**, covering all four findings. No test was skipped in either direction.
* `shellcheck -S warning` and `bash -n` clean on `survey.sh`.
* `check-scores-drift.py` (270 files) and `check-guide-scores.py` (56 guides) both exit 0. `docs/scores/`, `results.json`, `results.md` and `images.txt` are byte-identical to `origin/main`.
* PR body rewritten: it carried the corrected floor claim and did not describe A-D. A squash lands it as the commit message.

Not merging either PR. Ready for you — #683 then #684.

PR #684, opened 2026-08-01
ci(scores): put the weekly refresh's coverage in its PR body [IRO-727]

omerzamir:
The split half of IRO-727. **Gated by change-kind** (`.github/workflows/*`), so it is deliberately not riding along with #683 — and it should merge **after** #683, which writes the fields this reads.

## What is wrong today

The weekly rolling refresh PR says nothing about how much of the manifest the survey measured. It has been **256 of 295 rows** for months, and that PR is byte-indistinguishable from a full sweep. If tomorrow it drops 80 rows instead of 39, the PR still looks identical. The only trace is an Actions log that expires.

## Changes

* **Coverage in the PR body**, computed from the freshly written `results.json`: scanned / manifest rows, the skipped count broken down by stage, and a pointer to the `skipped` array that names each dropped row and why.
* A `results.json` that predates `schemaVersion` 1.1 reports **"not recorded"** rather than defaulting the missing fields to zero. My first draft defaulted them, and the four-arm test caught it printing *"256 of 256 — nothing was skipped"* against the current committed artifact, the exact shape whose 39 silent drops started this issue. Ironic and worth naming.
* Two stale **`~150 image`** comments corrected to the manifest's real size, and the PR body no longer claims "the top ~150 public images". (295 rows in `images.txt`, 256 currently scanned.)
* A note that `survey.sh` now fails the job on a coverage regression, so a degraded sweep cannot open a refresh PR at all.

## Not changed, deliberately

IRO-727 also flagged line 5's "generated from the **reproducible** isolation-survey" as part of the claim family IRO-723 retracted. **That premise does not hold** — re-derived against `main`: #671 touched `examples/isolation-survey/README.md` and left "reproducible" standing (README.md line 1 and line 9), and the word is still live in `docs/scores/index.md`, `docs/scores/leaderboard.md` and every `docs/scores/collections/*.md` footer. What #671 retracted was the *digest-pin* claim, not reproducibility. Retracting the word here alone would put this workflow out of step with a large CEO-ratified generated corpus, so it stays. If reproducibility should be retracted, that is a corpus-wide decision, not a one-file edit.

## Verification

The step's `run:` body was extracted with a parser and the coverage segment executed **unmodified** under `bash -euo pipefail` against four fixtures:

| fixture | output |
|---|---|
| committed schema-1.0 `results.json` | `256 scenarios scanned … dropped is **not recorded**` |
| 1.1 artifact with skips | `**256 of 295** … **2 skipped** (pull 1, run 1) …` |
| 1.1, full coverage | `**256 of 256** … Nothing was skipped.` |
| `results.json` missing | falls back to the error string, does not abort the step |

The whole extracted body passes `bash -n`; the file parses as YAML.

### DIFF
--- a/.github/workflows/scores-refresh.yml
+++ b/.github/workflows/scores-refresh.yml
@@ -39,8 +39,9 @@ jobs:
     # meaningless off the canonical repo.
     if: github.repository == 'IronSecCo/ironclaw'
     runs-on: ubuntu-latest
-    # The survey pulls + scans ~150 images through mirror.gcr.io; give it room but
-    # cap it so a hung pull can never burn a full 6h of runner time.
+    # The survey pulls + scans the 295 scenario rows in images.txt through
+    # mirror.gcr.io; give it room but cap it so a hung pull can never burn a full
+    # 6h of runner time.
     timeout-minutes: 120
     permissions:
       contents: write # push the rolling refresh branch
@@ -63,8 +64,13 @@ jobs:
         run: CGO_ENABLED=1 go build -o bin/ironctl ./cmd/ironctl
 
       # Run the full survey. PRUNE=1 removes each image right after it is scanned so
-      # the ~150-image sweep never exceeds the runner's free disk. Mirror-first
-      # (the default) dodges Docker Hub's anonymous pull-rate limit.
+      # the sweep never exceeds the runner's free disk. Mirror-first (the default)
+      # dodges Docker Hub's anonymous pull-rate limit.
+      #
+      # survey.sh exits non-zero when a scenario that scored in the committed
+      # results.json stops producing output (IRO-727), which fails this job before
+      # a degraded sweep can open a refresh PR. It still writes results.* first, so
+      # the run's artifacts name what was dropped even when it fails.
       - name: Run the isolation survey
         env:
           IRONCTL: ${{ github.workspace }}/bin/ironctl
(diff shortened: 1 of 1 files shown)


omerzamir:
Holding this one until #683 lands, not because anything here is wrong. I checked the four fixtures and the schema-1.0 arm is the right call: defaulting the missing fields to zero would have printed "nothing was skipped" against the exact artifact whose 39 silent drops opened IRO-727, so catching that in your own test was the important part.

Two confirmations so this does not get re-litigated when it comes up for merge:

- Keeping "reproducible" is correct and I am ratifying it. #671 retracted the digest-pin claim, not reproducibility, and the word is still live in `docs/scores/index.md`, `docs/scores/leaderboard.md` and every `docs/scores/collections/*.md` footer. A one-file edit here would put this workflow out of step with a CEO-ratified corpus. If reproducibility should go, that is a corpus-wide decision and it comes to me, not to this PR.
- The heredoc indentation is fine. The YAML block scalar strips the 10-space base indent, so `PY` lands at column 0 and the terminator resolves.

The one thing that will need a second look after #683: your comment says survey.sh "exits non-zero when a scenario that scored in the committed results.json stops producing output". #683 has a re-run laundering path that makes that not quite true locally, and I have asked for it to be fixed there. Once it is, this sentence holds as written and no change is needed here.

PR #685, opened 2026-08-01
Added a Rocket.Chat outbound channel adapter

### LINKED ISSUE #397: Add a Rocket.Chat outbound channel adapter
IronClaw ships outbound channel adapters for Slack, Discord, Teams, Mattermost, Telegram, Matrix, and more, but not Rocket.Chat, a popular self-hosted team chat. Rocket.Chat exposes a simple REST endpoint (`POST /api/v1/chat.postMessage`) with token headers, so it fits the existing `Adapter` interface cleanly. This is a self-contained, stdlib-only addition.

## What to do
Create `internal/host/channels/rocketchat.go` modeled on `internal/host/channels/mattermost.go`:
- Implement the `Adapter` interface (send an outbound message to a channel/room).
- POST to `{BaseURL}/api/v1/chat.postMessage` with `X-Auth-Token` and `X-User-Id` headers and a JSON body (`channel`, `text`).
- Make `BaseURL` overridable for tests (same pattern as the other adapters).
- Keep it stdlib-only (`net/http`, `encoding/json`), no new dependencies.

## File pointers
- Template to copy: `internal/host/channels/mattermost.go` (closest in shape) and its `mattermost_test.go`
- Interface + registration: `internal/host/channels/channels.go`
- Add `rocketchat_test.go` using an `httptest.Server` to assert the request path, headers, and body.

## Acceptance criteria
- `go test ./internal/host/channels/...` passes.
- Adapter is registered/selectable by its channel kind alongside the others.
- Test uses `httptest` and asserts the outbound request shape; no live network calls.

## Difficulty
Medium. The write-a-channel-adapter guide (`docs/writing-a-channel-adapter.md`) walks the interface end to end.

Phoenix1504e:
## What & why

Adds a stdlib-only Rocket.Chat channel adapter and registers it within the control plane daemon. This allows IronClaw to deliver outbound messages to Rocket.Chat channels via its REST API while maintaining strict secret redaction and zero external dependencies.

Closes #397 

## Changes

- Introduced `internal/host/channels/rocketchat.go` implementing `channels.Adapter` for Rocket.Chat using `POST /api/v1/chat.postMessage` with `X-Auth-Token` and `X-User-Id` header authentication.
- Added secret redaction helper logic to prevent exposing sensitive tokens or user IDs in log outputs or error messages.
- Added comprehensive unit tests in `internal/host/channels/rocketchat_test.go` covering message delivery, URL parameter overrides, and error redaction.
- Registered the Rocket.Chat adapter in `cmd/controlplane/main.go` gated behind the `IRONCLAW_ROCKETCHAT_BASE_URL` environment variable.

## Checklist

- [ ] **Docs updated:** README / `docs/**` reflect any user-facing or behavioral change.

## Validation

```bash
CGO_ENABLED=1 go test -v ./internal/host/channels/...
```

```text
=== RUN   TestRocketChatAdapter_Deliver_Success
--- PASS: TestRocketChatAdapter_Deliver_Success (0.00s)
=== RUN   TestRocketChatAdapter_RedactsSecretsInError
--- PASS: TestRocketChatAdapter_RedactsSecretsInError (0.00s)
PASS
ok      [github.com/IronSecCo/ironclaw/internal/host/channels](https://github.com/IronSecCo/ironclaw/internal/host/channels)    0.051s
```


Phoenix1504e:
@omerzamir can you please review this?

omerzamir:
Thanks for this, the shape matches the other adapters nicely. Two things before it can go in.

`rocketChatResponse.MessageID` is a `string` tagged `json:"message"`, but `chat.postMessage` returns `message` as an object, so against a real server the unmarshal always fails:

```
json: cannot unmarshal object into Go struct field rocketChatResponse.message of type string
```

Two consequences. The returned id is always the literal `"delivered"` instead of the real `_id`, and since the `success` check sits behind `err == nil` it never runs, so a 200 with `success:false` is reported as delivered. The test does not catch it because the fake server encodes the same struct back, so it only ever agrees with itself.

A nested `Message struct { ID string `json:"_id"` }` plus a test that writes a real captured response body would cover both.

Second, the four `IRONCLAW_ROCKETCHAT_*` vars are not added to `docs/channels.md`, `docs/env-reference.md`, or `deploy/helm/ironclaw/templates/secret.yaml`. Every other adapter is listed in all three, so as it stands the adapter is undiscoverable and unreachable from a Helm deploy.

Phoenix1504e:
@omerzamir updated the files with the suggestions, please check and approve

omerzamir:
My earlier comment was against a stale head, sorry. You had already fixed both of those in the Aug 29 commits, and the fixes are good: the nested `_id` struct is right, and the new test drives a raw response body with a 200 plus `success:false` case rather than encoding the struct back at itself. That is exactly the shape it needed.

Re-reviewed at 9ec1f1c. Two things left, one of them blocking.

The env var name does not match between the code and everything that documents it. `main.go` gates registration on `IRONCLAW_ROCKETCHAT_BASE_URL`, while `docs/channels.md`, `docs/env-reference.md` and the Helm secret template all say `IRONCLAW_ROCKETCHAT_URL`. Rendering the chart shows it:

```
$ helm template t deploy/helm/ironclaw --set secrets.rocketchatUrl=https://chat.example.com ...
  IRONCLAW_ROCKETCHAT_URL: "https://chat.example.com"
```

Since that var is the registration gate, a fully configured Helm deploy registers no adapter at all and says nothing about it. Worth picking one name and using it in all four places. Related: `values.yaml` has no `rocketchat*` keys, so there is nothing for a user to set there, unlike `signalCliUrl` and friends.

Second, the redaction test went away in the test refactor. The file header still promises the token and user id are redacted from errors, but nothing asserts it now. I replaced the whole body of `redact` with `return s` and the suite stayed green. Every other adapter in that package has a redaction test, so it would be good to keep yours.

Phoenix1504e:
@omerzamir Apologies for the env var mishap, changed it so its uniform across the board and added test to check if the keys are redacted or not from error messages

omerzamir (approved):
Re-checked at ab54dfd. Both points are resolved and I verified them rather than reading the diff:

Env vars line up end to end now. Rendering the chart with only the keys you added to values.yaml gives IRONCLAW_ROCKETCHAT_URL, which is exactly what main.go gates registration on, so the Helm path actually registers the adapter.

The redaction test is real. I replaced the whole body of redact with return s and it goes red on both assertions, so it is testing the thing it claims to.

Tests and build green locally, all required checks green. Approving.

omerzamir:
Merged, thanks Aditya. Good adapter, and the redaction test earns its keep.

One correction on my side: in my last comment I said I had reviewed a stale head and that you had fixed things before I asked. That was wrong, I misread the timeline. You responded to each round of review normally. Sorry for the confusion.

PR #694, opened 2026-08-10
feat(scan): add --list-dimensions flag to list scoring axes and weights

### LINKED ISSUE #446: Add 'ironctl scan --list-dimensions' to print scoring axes and weights

suprt:
## Summary

`ironctl scan` grades seven containment dimensions whose fixed weights sum to 100, but until now the only way to see the list and the weights was to read the `scorers` slice in `internal/host/scan/score.go`. This adds a `--list-dimensions` boolean flag that prints each dimension title and its weight and exits 0 without needing a container, so users understand what the score is made of before they run a scan.

The list is driven from the existing `scorers` slice through a new exported `Dimensions()` accessor (returning `DimensionWeight{Key, Title, Max}`), so adding a dimension in `score.go` later needs no second edit in `cmd/ironctl`. The internal `grade` funcs stay unexported — `Dimensions()` exposes only the published view.

## Changes

- **`internal/host/scan/score.go`** — new `DimensionWeight` type and `Dimensions()` accessor sitting next to `TotalWeight`; the public view of the internal `scorers` slice.
- **`cmd/ironctl/scan.go`** — `--list-dimensions` flag, an early return before the mode blocks (no-target, no-container path), a `printDimensions` helper that dot-pads labels to align the weight column, and a `scanUsage` help line.
- **`cmd/ironctl/scan_flags_test.go`** — `--list-dimensions` classified as image-safe (it carries no composite score) so `TestEveryScanFlagIsClassified` stays green.
- **`internal/host/scan/score_test.go`** — `TestDimensions` asserts the accessor stays in sync with `scorers` (same key/title/max, in order) and that the weights sum to `TotalWeight` (100).
- **`cmd/ironctl/scan_test.go`** — `TestCmdScan_ListDimensions` captures stdout, asserts `cmdScan --list-dimensions` exits nil, the header (`Containment dimensions (weights sum to 100)`) prints, and every dimension title from `scan.Dimensions()` appears (case-insensitive — output lowercases titles).

## Example output

```
$ ironctl scan --list-dimensions
Containment dimensions (weights sum to 100):
  non-root user (uid != 0).... 15
  dropped capabilities........ 20
  seccomp profile............. 15
  network isolation / egress.. 15
  read-only root filesystem... 10
  no docker.sock exposure..... 15
  no shared host namespaces... 10
```

## Verification

- `go build ./...` ✅
- `go test ./...` ✅ (all packages green, including new `TestDimensions`, `TestCmdScan_ListDimensions`, and `TestEveryScanFlagIsClassified`)
- `gofmt -l` on the touched files ✅ clean. (`gofmt -l .` does flag two pre-existing files on `main` — `cmd/scan-wasm/main.go` and `internal/host/scan/compare_test.go` — that are unrelated to this PR; I left them alone to keep the diff scoped.)

## Acceptance criteria

- [x] `ironctl scan --list-dimensions` prints all dimensions with weights and exits 0.
- [x] The list is derived from `scorers` (adding a dimension later needs no second edit) — via the `Dimensions()` accessor.
- [x] Flag documented in `scanUsage`.
- [x] Tests in `cmd/ironctl/scan_test.go` and `internal/host/scan` cover the listing.
- [x] `go build ./...`, `go test ./...`, `gofmt -l` on touched files clean.

Closes #446

### DIFF
--- a/cmd/ironctl/scan.go
+++ b/cmd/ironctl/scan.go
@@ -76,6 +76,7 @@ func cmdScan(args []string) error {
 	podmanBin := fs.String("podman-bin", envOrDefault("PODMAN", "podman"), "podman binary used for `podman container inspect` and `podman image inspect`")
 	nerdctlBin := fs.String("nerdctl-bin", envOrDefault("NERDCTL", "nerdctl"), "nerdctl binary used for `nerdctl container inspect` and `nerdctl image inspect`")
 	minScore := fs.Int("min-score", 0, "exit non-zero if the score is below this threshold (CI gate)")
+	listDims := fs.Bool("list-dimensions", false, "print the containment scoring dimensions and their weights, then exit")
 	fs.Usage = func() { scanUsage(os.Stdout) }
 	// Go's flag package stops at the first positional, so `scan <target> --json`
 	// would silently drop the flags after the target. Re-parse around each
@@ -92,6 +93,15 @@ func cmdScan(args []string) error {
 		}
 		positional = append(positional, rest[0])
 		rest = rest[1:]
+	}
+
+	// --list-dimensions: print the scoring axes and weights, then exit 0. This
+	// is a no-target, no-container path, so it returns BEFORE the mode blocks
+	// below. The list is sourced from scan.Dimensions() (the public view of the
+	// internal scorers slice), so adding a dimension in score.go later needs no
+	// second edit here.
+	if *listDims {
+		return printDimensions(os.Stdout)
 	}
 
 	// --compare A B: grade two live containers and print a side-by-side diff.
@@ -3172,6 +3182,28 @@ func podmanRootless(bin string) scan.Tristate {
 	}
 }
 
+// printDimensions prints the containment scoring axes with their fixed weights
+// (summing to scan.TotalWeight) and returns nil so cmdScan can surface it as a
+// clean exit 0. Titles come from scan.Dimensions(), so this stays in sync with
+// the scorers slice without a second source of truth. The label column is
+// left-justified to the longest title plus a small gap so the weights align.
+func printDimensions(w io.Writer) error {
+	dims := scan.Dimensions()
+	width := 0
+	for _, d := range dims {
+		if len(d.Title) > width {
+			width = len(d.Title)
+		}
+	}
+	fmt.Fprintf(w, "Containment dimensions (weights sum to %d):\n", scan.TotalWeight)
+	for _, d := range dims {
+		// width+4 pads the label to the longest title plus a 4-char gap before
+		// the weight, giving aligned columns without dots-fillers.
+		fmt.Fprintf(w, "  %-*s %d\n", width+4, d.Title, d.Max)
+	}
+	return nil
+}
+
 func scanUsage(w *os.File) {
 	fmt.Fprint(w, `ironctl scan — grade a container's containment posture (0-100)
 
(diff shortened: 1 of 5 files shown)


suprt:
Hi @omerzamir , can you check PR please?

omerzamir:
Nice flag, and I like that `Dimensions()` derives from the scorers slice so nothing needs a second edit later. Output looks right locally.

One thing to fix first. `captureStdout` says it "restores the original stdout even on failure", but there is no defer, so when `fn` calls `t.Fatalf` the restore never runs. I checked with a copy of the helper: the test after it still has `os.Stdout` pointing at the pipe, with nothing draining it, which will hang once the pipe buffer fills.

```go
os.Stdout = w
defer func() { os.Stdout = orig }()
```

right after the assignment covers it, and the `w.Close()` wants the same treatment. Happy to merge once that is in.

suprt:
Apologies, I missed this cleanup issue in the previous update, and thank you for catching it during review. `captureStdout` now restores the original `os.Stdout` and closes the pipe writer through deferred cleanup, so an early `t.Fatal` cannot leak the pipe into subsequent tests. The writer is also closed explicitly before waiting for the reader on the successful path, avoiding a deadlock.

omerzamir (approved):
My earlier comment was against a stale head, sorry about that. You had already fixed this in f3e530f two days before I wrote it.

Re-checked at f3e530f: I pulled your `captureStdout` out of the file verbatim and drove it with an aborting `fn`, and stdout is restored correctly now. Tests and build are green locally.

Approving.

omerzamir:
Merged, thanks Egor. Sourcing the list from the scorers slice was the right call, the output reads well.

One optional nit for next time: on the abort path the reader goroutine still blocks forever on `done <- string(buf)`, one leaked goroutine per aborted capture. A buffered `make(chan string, 1)` and a `defer r.Close()` would close it out.

omerzamir:
Small correction to my earlier comment here: I said I had reviewed a stale head and that you fixed this before I raised it. That was wrong, I misread the dates. The review came first and you fixed it. Nothing changes about the merge, just did not want the record to be misleading.

Issue #497, opened 2026-07-15
Write a container hardening guide for NATS


omerzamir:
We publish per-image hardening guides under `docs/blog/harden-<service>-container-isolation.md`. NATS is a popular lightweight messaging server and a good next guide.

**Scope (one new markdown file):**
- Run `ironctl scan` on the stock `nats` image; record the real stock grade + failing dimensions.
- Show the hardened stanza and improved grade. NATS is a broker - clients publish and subscribe over the network - so the honest ceiling is **89/100 grade B** (network held at WARN). Say so; do not claim A.
- Front-matter `title` + `description`; link to `../scores/nats.md`.
- Wire into `docs/blog/.nav.yml`, `docs/blog/index.md`, the `hardening-guides.md` hub (grade-B table), and the README list.

**Template:** copy `docs/blog/harden-rabbitmq-container-isolation.md` (same broker pattern).

**Acceptance:** `mkdocs build --strict` green; no em/en-dashes; numbers from a real scan.

PR #699, opened 2026-08-14
docs: add container hardening guide for NATS (#497)

PrinceBadru:
## What & why
- Write a container hardening guide for NATS  
- issue link :  https://github.com/IronSecCo/ironclaw/issues/497 

Closes #

## Changes
- Added container hardening guide for NATS based on real ironctl scan results (48/100, Grade D stock score). Honest ceiling is set to Grade B (89/100) due to network exposure requirements. 
- Ran mkdocs build --strict with zero errors.

## Checklist

- [ ] **Tests pass:** `CGO_ENABLED=1 go test ./...` is green (CGO is required — SQLCipher binding).
- [ ] **Formatted & vetted:** `gofmt -l .` is empty and `go vet ./...` passes.
- [ ] **Frozen contract:** `internal/contract/**` is **untouched** — or this carries an approved joint RFC (`docs/contract.md`) with code-owner sign-off.
- [ ] **Threat model:** no change to the sandbox seal / `network=none` / approval-gateway posture — or it is called out and reviewed.
- [ ] **Docs updated:** README / `docs/**` reflect any user-facing or behavioral change.
- [ ] **No secrets:** no tokens, keys, or credentials in code, tests, fixtures, or logs.
- [ ] **CLA signed:** first PR here? One click at <https://cla-assistant.io/IronSecCo/ironclaw>.

## Validation

```bash
CGO_ENABLED=1 go test ./...
```


omerzamir:
Thanks for picking up NATS. This one needs a rethink before it can land, because `docs/blog/harden-nats-container-isolation.md` already exists on main and the diff replaces it with a shorter version, so the net effect is a removal rather than a new guide.

Concrete issues in the replacement:

- The Compose block has no code fence. The file has `yaml` on a bare line followed by raw YAML, so MkDocs renders it as a paragraph.
- The image changes from `nats:2.10-alpine` to `nats:latest`, but `docs/scores/nats.md` is generated from the survey and grades `nats:2.10-alpine` at a pinned digest. The generated scorecard is the source of truth, so the guide has to match it.
- The frontmatter description promises "a full 100/100 grade A" while the body and the hub row say 89/100 B. 89 is the correct number: a broker has to accept connections, so the network dimension holds at a WARN by design.
- The row added to `hardening-guides.md` does not match that table's columns, which are Service, Default, Hardened, Why 89 is the ceiling.
- The README entry is missing its trailing comma and links a repo path while every neighbour links the published URL.
- No newline at end of file.

Worth saying: NATS really is missing from the README and from `docs/blog/index.md` on main. Adding it to those two lists, pointing at the guide that is already there, would be a genuinely useful small PR and I would take that one straight away.

